PR #38, opened 2020-01-02
[2338] Add results page

faissalMT:
### Context
In order to rewrite and restructure https://github.com/DFE-Digital/search-and-compare-ui a key piece of functionality is the results page.

Rewrite the results page in Rails as an API client for Manage Courses Backend.  

### Guidance to review
https://find-postgraduate-teacher-training.education.gov.uk/results

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


faissalMT:
[Build results page](https://trello.com/c/wH85UGvo/2338-build-results-page)

spec/decorators/course_decorator_spec.rb
```diff
@@ -374,4 +378,44 @@
...
+              accredited_body?: false,
+              website: "www.acmescitt.com",
+              address1: "Building 64",
+              address2: "32 Copton Lane",
+              address3: "Bracknel",
+              address4: "Berkshire",
+              postcode: "NXT STP")
+      end
+
+      it "can return a full address" do
```
gpeng: `it "returns...` would be more consistent with the rest of the file here and 410

app/controllers/results_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ResultsController < ApplicationController
+  def index
+    @course = Course
```
gpeng: Should this be `@courses`? Also, could this go in a query object or a scope on `Course`. This controller only really passes in pagination params but it needs to know a lot of relationships if we build the query in here.

spec/site_prism/page_objects/page/results.rb
```diff
@@ -0,0 +1,23 @@
...
+      element :provider_name, '[data-qa="course__provider_name"]'
+      element :description, '[data-qa="course__description"]'
+      element :name, '[data-qa="course__name"]'
+      element :accrediting_provider, '[data-qa="course__accrediting_provider"]'
+      element :funding_options, '[data-qa="course__funding_options"]'
+      element :main_address, '[data-qa="course__main_address"]'
+    end
+
+    class Results < SitePrism::Page
+      #def load_with_course(course)
```
gpeng: Can we delete this?

app/controllers/results_controller.rb
faissalMT: Since we're requesting this data over the wire (via JSON:API) it's better that we explicitly include relationships that we want.

gpeng: Ah right. Ok 👍

app/controllers/results_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ResultsController < ApplicationController
+  def index
+    @courses = Course
+      .includes(:provider)
+      .includes(:accrediting_provider)
+      .where(recruitment_cycle_year: Settings.current_cycle)
+      .where(provider_code: nil)
+      .page(params[:page])
+      .per(10)
```
misaka: I don't know how this has been done already in this app, but shouldn't this be parameterised?

spec/features/courses/results_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+  end
+
+  let(:accrediting_provider) { build(:provider) }
+  let(:decorated_course) { course.decorate }
+  let(:courses) { [course] }
+  let(:courses_request) do
+    stub_api_v3_request(
+      "/recruitment_cycles/#{Settings.current_cycle}/providers/courses?include=provider,accrediting_provider&page%5Bpage%5D=1&page%5Bper_page%5D=10",
+      resource_list_to_jsonapi(courses, include: %i[provider accrediting_provider]),
+    )
```
misaka: How about a `stub_api_v3_resource` based off of [`stub_api_v2_resource`](https://github.com/DFE-Digital/manage-courses-frontend/blob/de7378a5adc3d442db322c8aea6ceb267c37f7b9/spec/support/helpers.rb#L96-L114)?

spec/support/resource_list_to_jstonapi.rb
```diff
@@ -16,6 +16,7 @@ def resource_list_to_jsonapi(resource_list, **opts)
     },
     include: opts[:include],
     meta: opts[:meta],
+    links: opts[:links],
```
misaka: I know we're just adding one line here, but surely we should follow the boy-scout principle and fix the name of this file, no? ;)

app/controllers/results_controller.rb
misaka: Kaminari provider [a way](https://github.com/kaminari/kaminari#paginating-a-generic-array-object) to take a generic `Array` object and convert it into a paginated thing that can be used with Kaminari. 

I think that would give us a lot of pagination-related functionality for free, including stuff like a standardised view template. Would that be worth it?

spec/support/resource_list_to_jstonapi.rb
faissalMT: Didn't even notice

spec/features/courses/results_spec.rb
faissalMT: I think the `stub_api_v2_resource()` series of methods was at the wrong level of abstraction and complicated certain things by making assumptions such as that:  
 - there would always be a resource to stub  
 - `include` was the only argument we'd use
 - we'd only ever be getting data  
 - we'd never need to validate the data being sent  

These have lead to the method being complicated and doing too much, which meant alot of time was spent trying to re-engineer it to meet our new needs. As such I think we should wait until we have a better understanding of how we've used this API before we try to make another helper method like it.

spec/features/courses/results_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+            provider: provider2,
+            provider_code: provider.provider_code,
+            recruitment_cycle: current_recruitment_cycle,
+            description: "QTS",
+            funding_type: "salary")
+    end
+    let(:courses) { [course1, course2] }
+
+    it "Lists all courses" do
+      expect(results_page.courses.first.name.text).to eq("#{courses.first.name} (#{courses.first.course_code})")
```
danielburnley: Feel like it's worth moving some of this out to a method/some out to variables - hard to see what this is doing assertions against.

Varaibles could be `first_course`/`second_course`
Method could be `expect_course_to_appear_on_results_page(first_course`)

spec/features/courses/results_spec.rb
```diff
@@ -0,0 +1,148 @@
+#https://github.com/JsonApiClient/json_api_client/tree/db890adf3d7175e91829d359336ab7c98ed90a34#paginating
```
danielburnley: :eyes:

app/decorators/course_decorator.rb
```diff
@@ -111,6 +111,20 @@ def funding_option
     end
   end
 
+  def main_address
```
dankmitchell: This does already exist on the provider...

faissalMT: While this should exist on the provider, the method that currently exists in the provider has the wrong formatting

spec/features/courses/results_spec.rb
misaka: A lot of that is by design, tbh. `stub_api_v2_resource` is meant to be opinionated, for the sake of readability and convenience. It's really just a bit of sugar around specific types of requests: resources that can be retrieved. It's not meant to be a multi-tool that can handle all requests to the API.

app/decorators/provider_decorator.rb
```diff
@@ -1,13 +1,29 @@
 class ProviderDecorator < Draper::Decorator
   delegate_all
 
+  def short_address
+    address_lines.join(", ")
+  end
+
   def full_address
-    [object.address1, object.address2, object.address3, object.address4, object.postcode].select(&:present?).join("<br> ").html_safe
+    address_lines.map { |line| ERB::Util.html_escape(line) }.join("<br> ").html_safe
```
misaka: Should this, and the other related fixes, be part of a separate PR?

faissalMT: I can if we really want to. This is the only case of it on Find though and because I'm making other changes to this file as a result of refactoring I thought I might as well put it here. The commit for it is still marked with the appropriate card.

misaka: Fine by me.

PR #43, opened 2020-01-09
[2343] Add filters for study type

danielburnley:
study-type-filters

### Context

We need to mirror the study type filter page on the C# application

- Add a new study type filter page
- Add error handling to filter page

### DIFF
--- a/app/controllers/filters/study_type_controller.rb
+++ b/app/controllers/filters/study_type_controller.rb
@@ -1,0 +1,20 @@
+module Filters
+  class StudyTypeController < ApplicationController
+    def new; end
+
+    def create
+      if !(filter_params[:fulltime] == "True" || filter_params[:parttime] == "True")
+        flash[:error] = "You must make at least one selection"
+        redirect_to studytype_path(filter_params)
+      else
+        redirect_to results_path(filter_params)
+      end
+    end
+
+  private
+
+    def filter_params
+      params.permit(:fulltime, :parttime)
+    end
+  end
+end
--- a/app/views/filters/study_type/new.html.erb
+++ b/app/views/filters/study_type/new.html.erb
@@ -1,0 +1,78 @@
+<%= content_for :page_title, "Filter by study type" %>
+<% content_for :before_content do %>
+  <a href="<%= results_path %>" class="govuk-back-link">Back to search results</a>
+<% end %>
+
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      There is a problem
+    </h2>
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <li>
+          <a href="#studytype-error"><%= flash[:error] %></a>
+        </li>
+      </ul>
+    </div>
+  </div>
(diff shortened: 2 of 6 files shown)


app/controllers/filters/study_type_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Filters
+  class StudyTypeController < ApplicationController
+    def index; end
```
gpeng: These actions are more correctly `new` and `create` I think. It's a bit annoying because we need to match the C# URLs so we have to fudge the routes a bit.

app/views/filters/study_type/index.html.erb
```diff
@@ -0,0 +1,78 @@
+<%= content_for :page_title, "Filter by study type" %>
+<% content_for :before_content do %>
+  <a href="<%= results_path %>" class="govuk-back-link">Back to search results</a>
```
gpeng: We need to pass the params back here. There is a helper for this too in `app/helpers/view_helper.rb`. We need to link to the referrer I think (and protect against an [open redirect](https://owasp.org/www-project-cheat-sheets/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html)) I can do it in mine and update here if we can hold off deploying until I'm done.

config/routes.rb
```diff
@@ -14,4 +14,9 @@
 
   get "/course/:provider_code/:course_code", to: "courses#show", as: "course"
   get "/results", to: "results#index", as: "results"
+
+  scope "/results/filter" do
```
gpeng: I think...

```suggestion
  scope module: "filters", path: "/results/filter" do
```
will save you having to `to: "filters/`

spec/features/courses/results_spec.rb
```diff
@@ -37,7 +37,7 @@ def jsonapi_site_status(name, study_mode, status)
 
   let(:current_recruitment_cycle) { build :recruitment_cycle }
 
-  let(:results_page) { PageObjects::Page::Results.new }
+  let(:results_page) { PageObjects::Page::ResultsPage.new }
```
gpeng: the `Page` on `ResultsPage` is redundant here. I think the original was better.

spec/site_prism/page_objects/page/results/filters/study_type_page.rb
```diff
@@ -0,0 +1,16 @@
+module PageObjects
+  module Page
+    module Results
```
gpeng: If we make this module `ResultFilters` and move it to `spec/site_prism/page_objects/page/result_filters/` we don't have to change `Result` -> `ResultPage` and all is well in the world 😃

app/views/filters/study_type/new.html.erb
```diff
@@ -0,0 +1,78 @@
...
+              <span class="govuk-visually-hidden">Error:</span> You must make at least one selection
+            </span>
+          <% end %>
+          <div class="govuk-checkboxes">
+            <div class="govuk-checkboxes__item">
+              <%=
+                form.check_box(
+                  :fulltime,
+                  { class: "govuk-checkboxes__input", data: { qa: 'full_time'}, checked: params[:fulltime] == "True" },
+                  "True",
```
gpeng: Is there a reason you're not using the default boolean values here?

danielburnley: We're keeping it in line with the existing Find application - which is `True` and `False` :cry:

defong: dont think its case sensitive.... its windows based after all

PR #48, opened 2020-01-15
Send all applicants to Apply

davidgisbey:
### Context
Apply now has logic in place to direct the appropriate users to
the UCAS Apply service.

-  All users should now see the big green apply button which directs them to the Apply service

### Guidance

**Do not merge as the implementation has not been deployed on Apply yet**

### DIFF
--- a/app/views/courses/_apply.html.erb
+++ b/app/views/courses/_apply.html.erb
@@ -6,35 +6,12 @@
   <h2 class="govuk-heading-l" id="section-apply">Apply</h2>
 
   <% if course.has_vacancies? %>
-    <% if course.provider.supporting_dfe_apply? %>
-      <p class="govuk-body">
-        <%= link_to "Apply for this course", "https://www.apply-for-teacher-training.education.gov.uk/candidate/apply?providerCode=#{course.provider.provider_code}&courseCode=#{course.course_code}", class: "govuk-button govuk-button--start" %>
-      </p>
-    <% else %>
-      <p class="govuk-body">
-        <%= link_to 'Apply on the UCAS website', 'https://2020.teachertraining.apply.ucas.com/apply/student/login.do', class: 'govuk-link' %>. You’ll need to register with UCAS before you can apply.
-      </p>
-      <p class="govuk-body">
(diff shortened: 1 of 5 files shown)


app/views/courses/_apply.html.erb
```diff
@@ -6,35 +6,9 @@
...
-          <li>
-            Training programme code:
-            <span class="govuk-!-display-block govuk-!-font-size-36 govuk-!-font-weight-bold"><%= course.course_code %></span>
-          </li>
-        </ul>
-      </div>
-
-    <% end %>
+    <p class="govuk-body">
+      <%= link_to "Apply for this course", "https://www.apply-for-teacher-training.education.gov.uk/candidate/apply?providerCode=#{course.provider.provider_code}&courseCode=#{course.course_code}", class: "govuk-button govuk-button--start" %>
```
defong: where the test for the most important bit....

davidgisbey: It's code that is already in the codebase. I would assume it's already tested.

defong: Then it would of broken the tests....

defong: should apply url always point to prod?

spec/features/courses/show_spec.rb
```diff
@@ -229,6 +229,10 @@
       expect(course_page).to have_locations_map
 
       expect(course_page).to have_course_advice
+
+      expect(course_page).to have_content("Apply for this course")
```
defong: can we test the actual url as that will be the cta that the user will be going to

davidgisbey: Done 👍

davidgisbey:
As to whether it should always point to production. I don't see why it wouldn't. Definitely open to a reason it shouldn't, but none spring to mind.

app/views/courses/_apply.html.erb
davidgisbey: As to whether it should always point to production. I don't see why it wouldn't. Definitely open to a reason it shouldn't, but none spring to mind.

PR #49, opened 2020-01-15
add missing properties to course model

misaka:
### Context

As stated in the JsonApiClient docs, defining the schema for a model is
optional. However, it helps us understand what properties are available, and I
had to do this as part of the investigation into how API v3 is being used, so I
thought it may be worth adding.

https://github.com/JsonApiClient/json_api_client#schema

Add a bunch of properties to the `Course` model. These are already being used throughout the app. Also, alphabetise them.

### Guidance to review

Tests pass, so this should be fine, right??! I did have to guess at the types for these, but they seemed sensible (to me).


misaka:
Tests are _not_ passing. This is going to be tech debt.

PR #52, opened 2020-01-21
2786 pass params through filter pages

gpeng:
### Context
We are migrating search and compare to this app. The results page provides filters that are set in an interstitial page before returning the user to the results. Existing filter params need to be persisted across the get -> post -> get through the fitler page.

The back link also has to preserve filter parameters.

* Add concern that removes extraneous parameters from the request
* Add partial to write parameters to the filter page form (including support for array format paramters)
* Extend the `gov_uk_backlink_to` method to allow the link text to be set.
* Make the existing three filters consistent in naming, directory structure etc

### Guidance to review

The `FilterParams` module got slightly more complicated than we'd hoped. `params.permit!.except` would have been easier as these params aren't used for mass assignment so it would be safe. Brakeman disagreed 😠


app/controllers/concerns/filter_parameters.rb
```diff
@@ -0,0 +1,12 @@
+module FilterParameters
+private
+
+  def filter_params
+    params_to_pass = params.reject do |param|
+      %w(authenticity_token utf8 action controller).include?(param)
+    end
+
+    allowed_arrays = params_to_pass.keys.each_with_object({}) { |key, hash| hash[key] = [] }
+    params.permit(*params_to_pass.keys, allowed_arrays)
```
faissalMT: Why not use `request.query_parameters`

faissalMT: Why not use `request.request_parameters`?

gpeng: We need to also add the params from the filter form in here so... query_params + form params. I think we have to use param in that case if we're going to reuse this for all filter pages.

gpeng: Totally misread that as`query_parameters` not `request_parameters`. Yes, that would be loads simpler.

spec/features/result_filters/funding_spec.rb
```diff
@@ -80,13 +85,44 @@
         end
       end
     end
-  end
 
-private
+    describe "QS parameters" do
+      before do
+        stub_results_page_request
```
faissalMT: Not for this PR but `stub_results_page_request` is very specific and we might need to keep an eye on its usage.

gpeng: Yeah... definitely. I was just thinking of it as something for this bit of work really until we have a functional results page. It will break when we implement filtering on that page so we can address it then.

app/controllers/concerns/filter_parameters.rb
gpeng: Fixed in 4372712. `utf8` was still in there. I've filtered it out just because it isn't there in the existing implementation and I don't know what search and compare will do with it. Probably nothing 🤷‍♂

PR #55, opened 2020-01-22
2344 add qualifications filter

gpeng:
### Context

Part of the rebuild find epic. This adds the qualification filter page.

* Add `/results/filter/qualification`
* Add a view object for qualification to handle the checkbox state logic and parameters
* Update `FilterParams` to strip `authenticity_token`. This hasn't come up in the tests as it doesn't seem to be passed by I spotted it when running the app.
* Update azure-pipelines.yml with some magic from @rizzkhan1 as CI was failing this PR.

### Guidance to review

The view object `ResultFilters::QualificationView` is a new pattern for the app. I've added it as it is cleaner and allows us to test the logic that would otherwise be in the view template. This reduces the number of feature specs we need to write to test all scenarios. Performance isn't an issue with the test suite here yet but I think reducing dependence on slow feature tests and replacing with unit tests where appropriate would help to keep it that way.

### DIFF
--- a/app/controllers/result_filters/qualification_controller.rb
+++ b/app/controllers/result_filters/qualification_controller.rb
@@ -1,0 +1,24 @@
+module ResultFilters
+  class QualificationController < ApplicationController
+    include FilterParameters
+
+    before_action :create_view
+
+    def new; end
+
+    def create
+      if @view.qualification_selected?
+        redirect_to results_path(filter_params)
+      else
+        flash[:error] = "Please choose at least one qualification"
+        redirect_to qualification_path(filter_params.merge(qualifications: "none"))
+      end
+    end
+
+  private
+
+    def create_view
+      @view = ResultFilters::QualificationView.new(params: params)
+    end
+  end
+end
--- a/app/view_objects/result_filters/qualification_view.rb
+++ b/app/view_objects/result_filters/qualification_view.rb
@@ -1,0 +1,35 @@
+module ResultFilters
+  class QualificationView
+    def initialize(params:)
+      @qualifications_parameter = params[:qualifications]
+    end
+
+    def qts_only_checked?
+      checked?("QtsOnly")
+    end
+
+    def pgde_pgce_with_qts_checked?
+      checked?("PgdePgceWithQts")
+    end
+
+    def other_checked?
+      checked?("Other")
+    end
+
+    def qualification_selected?
+      parameter_array.any?
+    end
+
+  private
+
+    attr_reader :qualifications_parameter
+
+    def parameter_array
+      qualifications_parameter.present? ? qualifications_parameter.split(",") : []
+    end
+
+    def checked?(param_value)
+      parameter_array.empty? || parameter_array.include?(param_value)
+    end
+  end
+end
--- a/app/views/result_filters/_error_message.html.erb
+++ b/app/views/result_filters/_error_message.html.erb
@@ -1,0 +1,14 @@
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      You’ll need to correct some information.
+    </h2>
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <li>
+          <a href="#<%=error_anchor_id%>"><%= flash[:error] %></a>
+        </li>
+      </ul>
+    </div>
+  </div>
+<% end %>
--- a/app/views/result_filters/qualification/new.html.erb
+++ b/app/views/result_filters/qualification/new.html.erb
@@ -1,0 +1,107 @@
+<%= content_for :page_title, "Filter by qualification" %>
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(results_path(request.query_parameters), "Back to search results") %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render partial: "result_filters/error_message", locals: { error_anchor_id: "qualification-error" } %>
+
+    <%= form_with url: qualification_path, method: :post do |form| %>
+      <%= render 'result_filters/hidden_fields', exclude_keys: ["qualifications"], form: form %>
+      <fieldset role="radiogroup" aria-required="true" class="govuk-fieldset">
+        <legend>
+          <h1 class="govuk-heading-xl">What you will get</h1>
+        </legend>
+
+        <div class="govuk-form-group <%= flash[:error] ? "govuk-form-group--error" : "" %>">
+          <% if flash[:error] %>
+            <span id="qualification-error" class="govuk-error-message">
+              <span class="govuk-visually-hidden">Error:</span> You must make at least one selection
+            </span>
+          <% end %>
+          <div class="govuk-checkboxes">
+            <div class="govuk-checkboxes__item">
+              <%=
+                form.check_box(
+                  :qualifications,
+                  { 
+                    class: "govuk-checkboxes__input", 
+                    multiple: true, 
+                    data: { qa: 'qts_only' }, 
+                    checked: @view.qts_only_checked? 
+                  },
+                  "QtsOnly", 
+                  false
+                )
+              %>
+              <%= form.label :qualifications, value: "QtsOnly", class: "govuk-label govuk-checkboxes__label" do %>
+                <span class="govuk-!-font-weight-bold govuk-!-margin-bottom-2 govuk-!-display-block">QTS only</span>
+                <p class="govuk-body">
+                  QTS (qualified teacher status) allows you to teach in state schools in England.
+                </p>
+                <p class="govuk-body">
+                  It may also allow you to teach in <%= link_to "the EU, the EEA and Switzerland", "https://www.gov.uk/eu-eea", class: "govuk-link" %>. You should check with the institution you’d like to teach in for confirmation. The status of QTS in these countries may change after the UK leaves the EU. Please check <%= link_to "GOV.UK", "https://www.gov.uk", class: "govuk-link" %> for updates.
+                </p>
+                <p class="govuk-body">
+                  To teach in the rest of the UK or internationally, you may wish to consider a ‘PGCE (or PGDE) with QTS’ qualification instead.
+                </p>
+              <% end %>
+            </div>
+
+            <div class="govuk-checkboxes__item">
+              <%=
+                form.check_box(
+                  :qualifications,
+                  { 
+                    class: "govuk-checkboxes__input", 
+                    multiple: true, 
+                    data: { qa: 'pgde_pgce_with_qts'}, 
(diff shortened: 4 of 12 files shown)


gpeng:
[qualifications filter: Add filter page](https://trello.com/c/08m3kPfS/2344-qualifications-filter-add-filter-page)

app/controllers/result_filters/qualification_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    before_action :create_view
+
+    def new; end
+
+    def create
+      if @view.qualification_selected?
+        redirect_to results_path(filter_params)
+      else
+        flash[:error] = "You must make at least one selection"
```
defong: thats not the current behaviour

app/views/result_filters/qualification/new.html.erb
```diff
@@ -0,0 +1,120 @@
...
+
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      There is a problem
+    </h2>
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <li>
+          <a href="#qualification-error"><%= flash[:error] %></a>
```
defong: should deep link to field

azure-pipelines.yml
```diff
@@ -14,8 +14,10 @@ variables:
 
 steps:
 - script: |
+    set -x
```
defong: why this here?

azure-pipelines.yml
```diff
@@ -14,8 +14,10 @@ variables:
...
     GIT_SHORT_SHA=$(echo $(Build.SourceVersion) | cut -c 1-7)
     IMAGE_NAME_WITH_TAG=$(IMAGE_NAME):$GIT_SHORT_SHA
+    set +x
```
defong: why this here?

spec/site_prism/page_objects/page/result_filters/qualification.rb
```diff
@@ -0,0 +1,16 @@
+module PageObjects
+  module Page
+    module ResultFilters
+      class Qualification < SitePrism::Page
+        set_url "/results/filter/qualification{?query*}"
+
+        element :back_link, '[data-qa="page-back"]'
+        element :error, '[data-qa="error"]'
+        element :qts_only, '[data-qa="qts_only"]'
+        element :pgce_with_qts, '[data-qa="pgde_pgce_with_qts"]'
```
defong:
```suggestion
        element :pgde_pgce_with_qts, '[data-qa="pgde_pgce_with_qts"]'
```

spec/view_objects/result_filters/qualification_view_spec.rb
```diff
@@ -0,0 +1,100 @@
+require_relative "../../../app/view_objects/result_filters/qualification_view"
```
defong: do we need it...

maybe something we can do in config?

app/views/result_filters/qualification/new.html.erb
gpeng: fixed in 2b578fe

spec/site_prism/page_objects/page/result_filters/qualification.rb
gpeng: Updated in 2b578fe

app/controllers/result_filters/qualification_controller.rb
gpeng: I've asked @jesseyuen to review on the card. I'll update as required.

azure-pipelines.yml
gpeng: Recommendation from @rizzkhan1 as CI was falling over on this PR. It prints a trace of commands which seems to have fixed CI.

rizzkhan1: To display the following commands in the CI logs when executing

azure-pipelines.yml
gpeng: To reset that ^.

rizzkhan1: To disable verbose set earlier as that would conflict with following echo commands used for injecting values in the environmental variables

azure-pipelines.yml
rizzkhan1: this set -x is optional

spec/view_objects/result_filters/qualification_view_spec.rb
gpeng: We usually use `require "rails_helper"` but that loads the whole app. The class under test here is just a plain ruby object and doesn't need the whole app. So when we run the tests just for this class they run much faster if we just require the class under test.

gpeng:
> https://find-postgraduate-teacher-training.education.gov.uk/results/filter/qualification?qualifications=QtsOnly,PgdePgceWithQts,Other
> 
> where did this come from `PgdePgceWithQtsChecked`

Good spot. I've fixed that.

app/controllers/result_filters/qualification_controller.rb
gpeng: Update from Jesse. This is the correct behaviour.

gpeng:
I've added another commit (7473b6a) to fix an issue @dankmitchell spotted. Details in the commit.

gpeng:
> https://www.find-postgraduate-teacher-training.service.gov.uk/results?qualifications=QtsOnly%2CPgdePgceWithQts%2COther

@defong / @dankmitchell I've added a fix for this. Rails will support them so I've just unencoded them in the redirect to C# for now.

azure-pipelines.yml
defong: weird...

app/views/result_filters/qualification/new.html.erb
```diff
@@ -0,0 +1,120 @@
+<%= content_for :page_title, "Filter by qualification" %>
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(results_path(request.query_parameters), "Back to search results") %>
+<% end %>
+
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      There is a problem
```
defong:
```suggestion
You’ll need to correct some information. 
```

image
If the behaviour is correct than language used is questionable and seems aggressive...

`You must` vs `Please make` at least one selection.

`There is a problem` tell user there is a problem and nothing more without `full stop`.
` You’ll need to correct some information.` Tell user to do something correctly...

Are we sure:
- [ ] The wording/text
- [ ] Full width red border/styling

Which ever wording we use, can we have a test for the exacting wording as the error/validation display mechanism going to be reused...

gpeng: Updated in 1758477.

Screenshot 2020-01-23 at 14 45 10

gpeng: I think one of the other filters will need updating too but I'll do that in a separate PR.

spec/features/result_filters/qualifications_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        filter_page.back_link.click
+
+        expect_page_to_be_displayed_with_query(
+          page: results_page,
+          expected_query_params: { "test" => "params" },
+        )
+      end
+    end
+
+    describe "de-selecting an option" do
```
defong: both me and @TobyRet having a hard time with this

describe "de-selecting an option" 
it "allows the user to select a qualification"

spec/view_objects/result_filters/qualification_view_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        it { is_expected.to eq(true) }
+      end
+
+      context "when qualifications is empty" do
+        let(:params) { { qualifications: "" } }
+        it { is_expected.to eq(true) }
+      end
+
+      context "when qualifications is nil" do
+        let(:params) { {} }
```
defong: did you meant 

`        let(:params) { { qualifications: nil } }`

app/views/result_filters/qualification/new.html.erb
```diff
@@ -0,0 +1,107 @@
...
+                  To teach further education you don’t need QTS. Instead you can study for a PGCE or PGDE without QTS.
+                </p>
+              <% end %>
+            </div>
+          </div>
+        </div>
+      </fieldset>
+
+      <div class="govuk-form-group">
+        <%= form.submit "Find courses", name: nil, class: "govuk-button", data: { qa: 'find_courses' } %>
```
defong:
```suggestion
        <%= form.submit "Find courses", class: "govuk-button", data: { qa: 'find_courses' } %>
```

do we actually need `name: nil`?

app/view_objects/result_filters/qualification_view.rb
```diff
@@ -0,0 +1,35 @@
...
+    def qualification_selected?
+      parameter_array.any?
+    end
+
+  private
+
+    attr_reader :qualifications_parameter
+
+    def parameter_array
+      qualifications_parameter.nil? ? [] : qualifications_parameter.split(",")
```
defong:
```suggestion
      qualifications_parameter.present? ? qualifications_parameter.split(",") :[]
```

app/views/result_filters/_error_message.html.erb
```diff
@@ -0,0 +1,14 @@
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      You'll need to correct some information.
```
defong:
```suggestion
      You&#8217;ll need to correct some information.
```

https://github.com/DFE-Digital/publish-teacher-training/blob/d3c021bde8b17287b5bce0c7a15014344ed044ba/app/views/courses/_related_sidebar.html.erb#L26

image

spec/view_objects/result_filters/qualification_view_spec.rb
gpeng: No. I meant, when the key doesn't exist. I'll change the description of the context.

app/views/result_filters/qualification/new.html.erb
gpeng: It avoids 'commit: true' being passed as a parameter it think.

spec/features/result_filters/qualifications_spec.rb
gpeng: Fixed

spec/view_objects/result_filters/qualification_view_spec.rb
gpeng: Fixed

app/view_objects/result_filters/qualification_view.rb
gpeng: Fixed

app/views/result_filters/_error_message.html.erb
gpeng: Fixed

PR #56, opened 2020-01-23
Add .tool-versions config file

TobyRet:
### Context
Quick fix. No Trello ticket for this.

- Adds .tool-versions config file for the 'asdf-vm' version manager - https://github.com/asdf-vm/asdf


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.6.1
```
dankmitchell: Should we add the node version too?

TobyRet: 👍

PR #57, opened 2020-01-23
[2340] Location filter (Across england)

danielburnley:
### Context

In order to rewrite find we need to recreate the existing [locations filter](https://find-postgraduate-teacher-training.education.gov.uk/results/filter/location)

As the other two filters require additional support, this PR is only for the easiest option (Across England).

- Add the location filter page
- Add support for the parameter `l`
- Remove unsupported parameters 
- Display an error when no option is selected

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -1,0 +1,13 @@
+module ResultFilters
+  class LocationController < ApplicationController
+    include FilterParameters
+
+    def new; end
+
+    def create
+      params_without_unsupported_location_params = filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+
+      redirect_to results_path(params_without_unsupported_location_params)
+    end
+  end
+end
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -5,9 +5,18 @@ module ResultFilters
     def new; end
 
     def create
-      params_without_unsupported_location_params = filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+      if(!params[:l])
+        flash[:error] = "Please choose an option"
+        redirect_to location_path(params_without_unsupported_location_params)
+      else
+        redirect_to results_path(params_without_unsupported_location_params)
+      end
+    end
 
-      redirect_to results_path(params_without_unsupported_location_params)
+  private
+
+    def params_without_unsupported_location_params
+      filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
     end
   end
 end
--- a/app/views/result_filters/location/new.html.erb
+++ b/app/views/result_filters/location/new.html.erb
@@ -3,6 +3,21 @@
   <%= govuk_back_link_to(results_path(request.query_parameters), "Back to search results") %>
 <% end %>
 
+<% if flash[:error] %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error" data-qa="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      There is a problem
+    </h2>
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <li>
+          <a href="#studytype-error"><%= flash[:error] %></a>
+        </li>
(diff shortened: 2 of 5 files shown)


danielburnley:
https://trello.com/c/zAHjNVrV/2340-s-location-filter-add-filter-page-across-england

app/views/result_filters/location/new.html.erb
```diff
@@ -0,0 +1,53 @@
+<%= content_for :page_title, "Find courses by location or by training provider" %>
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(results_path(request.query_parameters), "Back to search results") %>
+<% end %>
+
+<% if flash[:error] %>
```
dankmitchell: I feel like this should go into a partial now. We are using it in number of places. Not a blocker for this PR.

gpeng: I've added it as a partial in the qualification filter now. I'll update the others when it gets approved.

app/controllers/result_filters/location_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      else
+        redirect_to results_path(params_without_unsupported_location_params)
+      end
+    end
+
+  private
+
+    def params_without_unsupported_location_params
+      filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+    end
```
misaka: Would it be worth it adding support for this to `FilterParams`? For example, `FilterParams.filter_params` does this:

```ruby
    custom_params = request.request_parameters.reject do |param|
      param.in? %w(utf8 authenticity_token)
    end
```

So what about doing this:

```ruby
    custom_params = request.request_parameters.reject do |param|
      param.in? %w(utf8 authenticity_token) + unsupported_params
    end
```

Where `unsupported_params` is defined in this class as `%i[lat lng rad query loc lq]`?

Also, if we decide to change `FilterParams.filter_params`, we can be good boy-scouts and use `except` like the above.

```ruby
    custom_params = request.request_parameters.except(%i(utf8 authenticity_token) + unsupported_params)
```

spec/site_prism/page_objects/page/result_filters/location.rb
```diff
@@ -0,0 +1,14 @@
+module PageObjects
+  module Page
+    module ResultFilters
+      class Location < SitePrism::Page
+        set_url "/results/filter/location{?query*}"
```
misaka:
```suggestion
        set_url "/results/filter/location{?test%5B%5D*}"
```

Took a bit of digging to sort this out, but basically in the RFC for these URL templates, `query` here is an [`expression`](https://tools.ietf.org/html/rfc6570#section-2.2) which is constituted of a [`variable-list`](https://tools.ietf.org/html/rfc6570#section-2.3), which may include `pct-encoded` elements that, in the "[notation convention](https://tools.ietf.org/html/rfc6570#section-1.5)", escapes reserved characters (such as "[" and "]") using hex digits. Which means `test[]` has to be escaped as `test%5B%5D`. In my testing, this seems to work.

spec/features/result_filters/location_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+          "test" => "value",
+        },
+      )
+    end
+
+    it "passes arrays correctly" do
+      url_with_array_params = "#{filter_page.url}?test[]=1&test[]=2"
+      PageObjects::Page::ResultFilters::Location.set_url(url_with_array_params)
+
+      filter_page.load
```
misaka:
```suggestion
      filter_page.load("test%5B%5D" => [1, 2])
```

And with the changes mentioned below in the SitePrism page, we shouldn't need the two lines above (149 and 150 currently).

spec/site_prism/page_objects/page/result_filters/location.rb
danielburnley: `query` is used in the tests to load pages with a specific query - although arrays can be a bit more finicky. But we `query` in all the other tests and changing it to this would break all of those

`page.load(query: { foo: 'bar' })` is an example of how we use this

app/controllers/result_filters/location_controller.rb
danielburnley: I have a feeling this code will change in the subsequent updates and be in this controller still - lat/lng/rad (and maybe a couple others) will be needed to change in the Postcode ticket, and `lq` in the provider name one will need to change in the provider ticket. So I'll hold off on changing this for now

PR #58, opened 2020-01-24
Make filter error messages consistent

gpeng:
### Context

We are replacing the search and compare UI filter pages. As they have been built concurrently there are some inconsistencies.

Make the error message consistent across filters -> 

* User the `result_filters/error_message` partial
* Link in to the error
* Render error messages in the 2/3 column of the page


gpeng:
[Make filter error messages consistent](https://trello.com/c/tjCsV48H/2854-make-filter-error-messages-consistent)

PR #59, opened 2020-01-27
[2848] Fix financial information for apprenticeships

dankmitchell:
### Context
Apprenticeships come with a salary so the information shown for these courses should match a standalone salaried course i.e.
https://www.find-postgraduate-teacher-training.service.gov.uk/course/2BZ/39NZ#section-financial-support

Return `salaried?` true for apprenticeships 

### Guidance to review
Should match -
https://www.find-postgraduate-teacher-training.service.gov.uk/course/2BZ/39NZ#section-financial-support

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -75,7 +75,7 @@ class CourseDecorator < Draper::Decorator
   end
 
   def salaried?
-    object.funding_type == "salary"
+    object.funding_type == "salary" || object.funding_type == "apprenticeship"
   end
 
   def apprenticeship?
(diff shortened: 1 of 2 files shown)


dankmitchell:
[Check financial incentives data](https://trello.com/c/mVrHhGra/2848-check-financial-incentives-data)

app/decorators/course_decorator.rb
```diff
@@ -75,7 +75,7 @@ def scholarship_amount
   end
 
   def salaried?
-    object.funding_type == "salary"
+    object.funding_type == "salary" || object.funding_type == "apprenticeship"
```
defong: https://github.com/DFE-Digital/teacher-training-api/blob/2fcb258955cab533f2845c67d3e9d85bf1c4546e/app/models/course.rb#L311

Should it not be using this instead?

dankmitchell: Possibly but I copied what search-and-compare-ui does which is the same as the preview in publish. Possible refactor post-launch for both preview and show

PR #60, opened 2020-01-27
2853 add links for filter pages with default filtering params

gpeng:
### Context

We are rebuilding the results page. The links to the filter pages on the results page need to pass the querystring into the filter pages but also populate some default values if they aren't present.

* Add `ResultsView` to handle this logic
* Add links for each of the filters. The filter boxes themselves are out of scope for this work. There aren't any feature specs in this PR for the links as they can be added to the filter feature specs once the full feature 'journey' (ie visit results, click filter, fill filter, return to results, check courses returned) is present to test.
* Update the azure config again as the fix we put in seems to no longer be required and now actually breaks CI. 

Screenshot 2020-01-27 at 15 45 25

### Guidance to review

The results page redirects to `localhost:5000` in development by default. Temporarily update `config/settings.yml` to test.

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -1,0 +1,42 @@
+class ResultsView
+  def initialize(query_parameters:)
+    @query_parameters = query_parameters
+  end
+
+  def query_parameters_with_defaults
+    query_parameters.except("utf8", "authenticity_token")
+      .merge(qualifications_parameters)
+      .merge(fulltime_parameters)
+      .merge(parttime_parameters)
+      .merge(hasvacancies_parameters)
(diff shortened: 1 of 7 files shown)


gpeng:
[[M] Add links for filter pages with default filtering params](https://trello.com/c/W3pwOA7f/2853-m-add-links-for-filter-pages-with-default-filtering-params)

app/view_objects/results_view.rb
```diff
@@ -0,0 +1,42 @@
+class ResultsView
```
danielburnley: This is a neat pattern :eyes:

PR #62, opened 2020-01-29
[2783] - Filter courses by location

TobyRet:
### Context
Adds an option to the `/results/filter/location` to allow users to filter requests by location.

- Upon submitting the form, the location query is geocoded and the params are populated with additional key/values (:lq, :lat, :lng)
- If a user's location query returns no results from the Google API then the user is redirected back to the page and an error displays
- Same deal if no option is selected

#### 1. Unknown location
error

#### 2. Known location
success

### Guidance to review
- After some feedback on the approach taken in certain areas. See comments below

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -7,13 +7,19 @@ module ResultFilters
     def new; end
 
     def create
-      if(!params[:l])
-        flash[:error] = "Please choose an option"
-        redirect_to location_path(params_without_unsupported_location_params)
-      elsif params[:l] == "3"
+      if params[:l] == "3"
         redirect_to provider_path(params_for_provider_search)
+      end
+
+      form_params = strip(filter_params.clone)
+      form_object = LocationFilterForm.new(form_params)
+
+      if form_object.valid?
+        all_params = form_params.merge!(form_object.params)
+        redirect_to results_path(all_params)
       else
-        redirect_to results_path(params_without_unsupported_location_params)
+        flash[:error] = form_object.errors
+        redirect_to location_path(form_params)
       end
     end
 
@@ -28,8 +34,8 @@ module ResultFilters
       filter_params.except(:lat, :lng, :rad, :loc, :lq)
     end
 
-    def params_without_unsupported_location_params
-      filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+    def strip(params)
+      params.reject { |_, v| v == "" }
     end
   end
 end
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -7,7 +7,7 @@ module ResultFilters
     def new; end
 
     def create
-      if params[:l] == "3"
+      if provider_option_selected?
         redirect_to(provider_path(params_for_provider_search)) && return
       end
 
@@ -28,6 +28,10 @@ module ResultFilters
     def build_results_filter_query_parameters
       @results_filter_query_parameters = ResultsView.new(query_parameters: request.query_parameters)
         .query_parameters_with_defaults
+    end
+
+    def provider_option_selected?
+      filter_params[:l] == "3"
     end
 
     def params_for_provider_search
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -8,7 +8,7 @@ module ResultFilters
 
     def create
       if params[:l] == "3"
-        redirect_to provider_path(params_for_provider_search)
+        redirect_to(provider_path(params_for_provider_search)) && return
       end
 
       form_params = strip(filter_params.clone)
--- a/app/form_objects/result_filters/location_filter_form.rb
+++ b/app/form_objects/result_filters/location_filter_form.rb
@@ -7,12 +7,12 @@ module ResultFilters
 
     def initialize(params)
       @params = params
-      @errors = nil
+      @errors = []
     end
 
     def valid?
       validate
-      @errors.nil?
+      @errors.empty?
     end
 
   private
@@ -20,10 +20,10 @@ module ResultFilters
     def validate
       case selected_option
       when NO_OPTION
-        handle_no_option
+        @errors = ["Please choose an option"]
       when LOCATION_OPTION
         if location_query.nil?
-          handle_missing_location
+          @errors = ["Postcode, town, or city", "Please enter a postcode, city or town in England"]
         else
           handle_location_option
         end
@@ -36,16 +36,8 @@ module ResultFilters
         @params.merge!(geocode_params)
         @valid = true
       else
-        @errors = "We couldn't find this location, please check your input and try again"
+        @errors = ["Postcode, town or city", "We couldn't find this location, please check your input and try again"]
       end
-    end
-
-    def handle_missing_location
-      @errors = "Please enter a postcode, town or city in England"
-    end
-
-    def handle_no_option
-      @errors = "Please choose an option"
     end
 
     def geocode_params_for(query)
--- a/app/form_objects/result_filters/location_filter_form.rb
+++ b/app/form_objects/result_filters/location_filter_form.rb
@@ -1,0 +1,76 @@
+module ResultFilters
+  class LocationFilterForm
+    NO_OPTION = nil
+    LOCATION_OPTION = "1".freeze
+
+    attr_reader :params, :errors
+
+    def initialize(params)
+      @params = params
+      @errors = nil
+    end
+
+    def valid?
+      validate
+      @errors.nil?
+    end
+
+  private
+
+    def validate
+      case selected_option
+      when NO_OPTION
+        handle_no_option
+      when LOCATION_OPTION
+        if location_query.nil?
+          handle_missing_location
+        else
+          handle_location_option
+        end
+      end
+    end
+
+    def handle_location_option
+      geocode_params = geocode_params_for(location_query)
+      if geocode_params
+        @params.merge!(geocode_params)
+        @valid = true
+      else
+        @errors = "We couldn't find this location, please check your input and try again"
+      end
+    end
+
+    def handle_missing_location
+      @errors = "Please enter a postcode, town or city in England"
+    end
+
+    def handle_no_option
+      @errors = "Please choose an option"
+    end
+
+    def geocode_params_for(query)
+      results = Geocoder.search(query, components: "country:UK")
+      england_results = results.select { |r| r.state == "England" }.first
+      if england_results
+        {
+            lat: england_results.latitude,
+            lng: england_results.longitude,
+            loc: england_results.address,
+            lq: location_query,
+        }
+      end
+    end
+
+    def selected_option
+      @params[:l]
+    end
+
+    def location_query
+      @params[:lq]
+    end
+
+    def search_radius
+      @params[:rad]
+    end
+  end
+end
--- a/app/form_objects/result_filters/location_filter_form.rb
+++ b/app/form_objects/result_filters/location_filter_form.rb
@@ -20,10 +20,10 @@ module ResultFilters
     def validate
       case selected_option
       when NO_OPTION
-        @errors = ["Please choose an option"]
+        @errors = [I18n.t("location_filter.errors.no_option")]
       when LOCATION_OPTION
         if location_query.nil?
-          @errors = ["Postcode, town, or city", "Please enter a postcode, city or town in England"]
+          @errors = [I18n.t("location_filter.fields.location"), I18n.t("location_filter.errors.missing_location")]
         else
           handle_location_option
         end
@@ -36,7 +36,7 @@ module ResultFilters
         @params.merge!(geocode_params)
         @valid = true
       else
-        @errors = ["Postcode, town or city", "We couldn't find this location, please check your input and try again"]
+        @errors = [I18n.t("location_filter.fields.location"), I18n.t("location_filter.errors.unknown_location")]
       end
     end
 
--- a/app/views/result_filters/location/new.html.erb
+++ b/app/views/result_filters/location/new.html.erb
@@ -8,11 +8,11 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render partial: "result_filters/error_message", locals: {
       error_anchor_id:
-        if flash[:error] && flash[:error].include?("Please choose an option")
+        if no_option_selected?
           "search-options"
-        elsif flash[:error] && flash[:error].include?("Postcode, town, or city")
+        elsif location_error?
           "location-error"
-        elsif flash[:error] && flash[:error].include?("Training provider")
+        elsif provider_error?
           "provider-error"
         end
     } %>
@@ -38,13 +38,13 @@
               <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
             </div>
             <div
-              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless flash[:error] && flash[:error].include?("Please enter a postcode, city or town in England") %>"
+              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? %>"
               id="location-conditional">
               <div class="govuk-form-group">
                 <%= form.label :lq, "Postcode, town or city", class: "govuk-label" %>
-                <% if flash[:error] && flash[:error].count > 1 %>
+                <% if location_error? %>
                   <div class="govuk-error-message" id="location-error" data-qa="location-error">
-                    <span class="govuk-visually-hidden">Error:</span><%= flash[:error].last  %>
+                    <span class="govuk-visually-hidden">Error: </span><%= flash[:error].last  %>
                   </div>
                 <% end %>
                 <%=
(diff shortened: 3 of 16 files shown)


app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,70 @@ module ResultFilters
...
+    def strip(params)
+      params.reject { |_, v| v == "" }
+    end
+
+    def geocode_params_for(query)
+      # Google's Geocoding API is not always reliable
+      # https://stackoverflow.com/questions/2647086/googles-geocoder-returns-wrong-country-ignoring-the-region-hint
+      # e.g. A query 'FOOBAR' still gets a hit from the Netherlands even if
+      # the region is restricted by 'uk'.
+      # Appending 'uk' to the query appears to prevent this from happening.
```
TobyRet: Or an alternative approach would be to *not* append 'uk' and instead select / filter only those hits that are from the 'uk'. Anyone have a preference?

TobyRet:
[[L] location filter: Add filter page - by location](https://trello.com/c/qeSWGWrJ/2783-l-location-filter-add-filter-page-by-location)

config/settings.yml
```diff
@@ -9,3 +9,4 @@ service_support:
 search_and_compare_ui:
   base_url: http://localhost:5000
 redirect_results_to_c_sharp: true
+gcp_api_key: please_change_me
```
gpeng: `please_commit_me` 🤣

app/controllers/result_filters/location_controller.rb
misaka: In fact, we should only be dealing with England, I thought. Should we limit it to that, even?

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,70 @@ module ResultFilters
...
+        redirect_to results_path(build_params)
       end
     end
 
   private
 
-    def params_without_unsupported_location_params
-      filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+    def handle_no_option(build_params)
+      flash[:error] = "Please choose an option"
```
gpeng: We've used `app/views/result_filters/_error_message.html.erb` in the other filters to make copy and formatting consistent.

app/controllers/result_filters/location_controller.rb
gpeng: Maybe filtering the results (or maybe both) would protect us more if we get strange results. There's no guarantee that adding UK to the query will always return the results we want. That said, it's highly unlikely I suppose and if it adds a lot of complexity then 🤷‍♂

config/settings.yml
gpeng: `please_commit_me` 🤣

app/view_objects/result_filters/location_view.rb
```diff
@@ -0,0 +1,23 @@
+module ResultFilters
+  class LocationView
```
misaka: They way this is implemented, it's closer to a form object than a view object. A form object is meant to abstract logic that's relevant to inputs sent into the app from a form. For example, we could've used that for the create wizard to have validations that are specific to the way the form is used (we could've used that for the modern language subjects issue we looked at recently).

A view object is for logic that is specific to the view, instead of using helpers, for example. We mentioned presenters also, those usually wrap models in a way to make them presentable in views (aka decorators, we already use these via the gem `draper` in publish).

I think the confusion comes partially from the way `ResultsFilter::QualificationView` is implemented, tbh (sorry @gpeng :) ). It blurs the lines a little bit by providing helper-style functionality that inspects the params.

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,70 @@ module ResultFilters
...
+      redirect_to location_path(build_params)
+    end
+
+    def handle_location_option(build_params)
+      geocode_params = geocode_params_for(@view.location_query)
```
defong: image

What happened to this?

app/controllers/result_filters/location_controller.rb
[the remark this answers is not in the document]
TobyRet: 👌

app/controllers/result_filters/location_controller.rb
TobyRet: I missed that edge case sorry. Will add it in 👍

app/controllers/result_filters/location_controller.rb
TobyRet: OK no worries - found a solution that restricts results to England only. Just requires tweaking the params.

```
irb(main):041:0> Geocoder.search('Edinburgh', components: 'country:england')
=> []
irb(main):042:0> Geocoder.search('FOOBAR', components: 'country:england')
=> []
```

app/controllers/result_filters/location_controller.rb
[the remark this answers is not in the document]
TobyRet: 👍

config/settings.yml
TobyRet: Boo 😞

app/controllers/result_filters/location_controller.rb
TobyRet: Looks like a tweak of the params fixes the issue

```
 Geocoder.search('FOOBAR', components: 'country:UK')
=> [#<Geocoder::Result::Google:0x00007fb41f8e1068 @data={"address_components"=>[{"long_name"=>"1", "short_name"=>"1", "types"=>["street_number"]}, {"long_name"=>"Rathbone Square", "short_name"=>"Rathbone Square", "types"=>["route"]}, {"long_name"=>"Fitzrovia", "short_name"=>"Fitzrovia", "types"=>["neighborhood", "political"]}, {"long_name"=>"London", "short_name"=>"London", "types"=>["postal_town"]}, {"long_name"=>"Greater London", "short_name"=>"Greater London", "types"=>["administrative_area_level_2", "political"]}, {"long_name"=>"England", "short_name"=>"England", "types"=>["administrative_area_level_1", "political"]}, {"long_name"=>"United Kingdom", "short_name"=>"GB", "types"=>["country", "political"]}, {"long_name"=>"W1T 1FB", "short_name"=>"W1T 1FB", "types"=>["postal_code"]}], "formatted_address"=>"1 Rathbone Square, Fitzrovia, London W1T 1FB, UK", "geometry"=>{"location"=>{"lat"=>51.5167849, "lng"=>-0.1340803}, "location_type"=>"ROOFTOP", "viewport"=>{"northeast"=>{"lat"=>51.5181338802915, "lng"=>-0.132731319708498}, "southwest"=>{"lat"=>51.5154359197085, "lng"=>-0.135429280291502}}}, "place_id"=>"ChIJUXyR96obdkgRUs5NVnKio00", "plus_code"=>{"compound_code"=>"GV88+P9 London, United Kingdom", "global_code"=>"9C3XGV88+P9"}, "types"=>["bar", "establishment", "point_of_interest"]}, @cache_hit=nil>]
```

TobyRet: Doesn't look like the search can be further refined to England though

TobyRet:
Display error when user selects location option but does not enter postcode

no_Postcode

TobyRet:
Allow user to select search radius

rad

app/view_objects/result_filters/location_view.rb
TobyRet: I have removed this file and put the params helper methods back into the LocationController

app/controllers/result_filters/location_controller.rb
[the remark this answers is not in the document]
TobyRet: @gpeng 
Yep this has been implemented - https://github.com/DFE-Digital/find-teacher-training/blob/08d3cf15874eeff45183f84a73b34494de417c11/app/views/result_filters/location/new.html.erb#L8

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,89 @@ module ResultFilters
...
       end
     end
 
   private
 
-    def params_without_unsupported_location_params
-      filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
+    def handle_location_option(build_params)
+      geocode_params = geocode_params_for(location_query)
+      if !geocode_params
```
faissalMT: I'd do this statement the opposite way

TobyRet: Yep good spot. 👍 I've pushed the change

app/views/result_filters/location/new.html.erb
```diff
@@ -14,31 +14,76 @@
...
+                    :lq,
+                    value: "#{flash[:query]}",
+                    class: "govuk-input",
+                    data: { qa: "location-query" },
+                    )
+                %>
+              </div>
+              <div class="govuk-form-group">
+                <%= form.label :rad, "Within", class: "govuk-label" %>
+                <%= form.select :rad,
```
defong: 5, 10, 20 50, 100

app/views/result_filters/location/new.html.erb
```diff
@@ -14,31 +14,76 @@
...
+                  aria: { 'controls': "location-conditional" }
                 )
               %>
-              <%= form.label :l_2, "Across England", class: "govuk-label govuk-radios__label" %>
+              <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
+            </div>
+            <div class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless flash[:location_error] %>" id="location-conditional">
+              <div class="govuk-form-group">
+                <%= form.label :lq, "Postcode, town or city", class: "govuk-label" %>
+                <div class="govuk-error-message" data-qa="location-error"><%= flash[:location_error] %></div>
```
defong: `data-qa="location-error"`

that needs to be id

as clicking on the link

image

do not jump to section

TobyRet: Cool 👍 
Have pushed fix

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,89 @@ module ResultFilters
...
+      end
+    end
+
+    def strip(params)
+      params.reject { |_, v| v == "" }
+    end
+
+    def geocode_params_for(query)
+      results = Geocoder.search(query, components: "country:UK")
+      england_results = results.select { |r| r.state == "England" }.first
```
defong: https://github.com/DFE-Digital/search-and-compare-api/blob/3aa6be82dcf5548892580e900ea1e37decf63a0f/src/api/Services/Geocoder/TECH_DEBT__TemporarilyCopied__Geocoder.cs#L61-L65

Not too sure about tying it down to `England` only

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,89 @@ module ResultFilters
...
+
+    def build_errors_for(type)
+      flash[:query] = location_query
+
+      case type
+      when :none_selected
+        flash[:error] = "Please choose an option"
+      when :missing_location
+        flash[:error] = "Postcode, town or city"
+        flash[:location_error] = "Please enter a postcode, city or town in England"
```
defong:
```suggestion
        flash[:location_error] = "Please enter a postcode, town or city in England"
```

image

I think you ported c# bug along...

app/views/result_filters/location/new.html.erb
```diff
@@ -14,31 +14,76 @@
...
+                <%= form.select :rad,
+                                options_for_select(
+                                  [
+                                    ["5 miles", "5"],
+                                    ["10 miles", "10"],
+                                    ["20 miles", "20"],
+                                    ["50 miles", "50"],
+                                    ["100 miles", "100"],
+                                  ],
+                                  ["20"]
```
defong: will 20 be overriden by the user when they revisit to up the value?

ie

```
user tried 100 miles, comes back and down it to 50 (what value they see before change 20 or 100?)
```

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,89 @@ module ResultFilters
   class LocationController < ApplicationController
     include FilterParameters
 
+    NO_OPTION = nil
+    LOCATION_OPTION = "1".freeze
```
defong: available options 

https://github.com/DFE-Digital/search-and-compare-ui/blob/37c4385ea5764a05743078d23012122524e45423/src/Filters/Enums/LocationOption.cs

app/controllers/result_filters/location_controller.rb
TobyRet: But the error messages / guidance say to enter a `postcode in, city or town in England`

app/controllers/result_filters/location_controller.rb
TobyRet: This is about the value of the `l` parameter. If it's nil then no option was selected. If 1, `Across England` was selected ...

There might be a need to revisit this once @faissalMT 's work is merged

spec/features/result_filters/location_spec.rb
```diff
@@ -23,110 +24,90 @@
...
-      expect_page_to_be_displayed_with_query(
-        page: results_page,
-        expected_query_params: {
-          "l" => "2",
-        },
-      )
+  describe "Navigating to the page with currently selected filters" do
+    it "Preselects across england" do
+      filter_page.load(query: { l: 2 })
+      expect(filter_page.across_england.checked?).to eq(true)
```
defong:
```suggestion
      expect(filter_page.across_england).to be_checked
```

app/controllers/result_filters/location_controller.rb
```diff
@@ -2,21 +2,92 @@ module ResultFilters
...
+
     def new; end
 
     def create
-      if(!params[:l])
-        flash[:error] = "Please choose an option"
-        redirect_to location_path(params_without_unsupported_location_params)
+      build_params = strip(filter_params.clone)
+
```
gpeng: This would be a good place for the form object pattern we've been discussing. The form object would be active model-esque and have `errors` and `valid?` so the controller would then end up being something like:

```ruby
form_object = LocationForm.new(params)
all_params = filter_params.merge(form_object.params)

if form_object.valid?
  redirect_to results_path(all_params)
else
  flash[:error] = form_object.errors
  redirect_to location_path(all_params)
end
```

TobyRet: Yep good idea. I'll make the change

app/controllers/result_filters/location_controller.rb
TobyRet: New bug ticket has been created for error messages

app/views/result_filters/location/new.html.erb
TobyRet: Fixed 👍

app/controllers/result_filters/location_controller.rb
TobyRet: Going to refactor in a separate PR

> Failing validation...
> no input and rad selected to 5 miles
> 
> image
> 
> issues
> 
> * [ ]  the back link got changed...
> * [x]  my previous selection of 5 miles got lost

New ticket for back link bug has been created

app/form_objects/result_filters/location_filter_form.rb
```diff
@@ -0,0 +1,82 @@
...
+      @valid = false
+    end
+
+    def handle_no_option
+      @errors = "Please choose an option"
+      @valid = false
+    end
+
+    def geocode_params_for(query)
+      results = Geocoder.search(query, components: "country:UK")
```
TobyRet: @gpeng - I have not extracted the geocoding to a separate service as it's only one line. Don't see much point unless you think otherwise ...

spec/features/result_filters/location_spec.rb
```diff
@@ -23,110 +24,81 @@
   describe "Selecting an option" do
     before { filter_page.load }
 
-    it "Allows the user to select across england" do
-      filter_page.across_england.click
-      filter_page.find_courses.click
+    it "Allows the user to select 'by postcode, town or city'" do
+      location_filter_results_page = PageObjects::Page::ResultFilters::LocationFilterResults.new
```
TobyRet: See commit message for explanation - https://github.com/DFE-Digital/find-teacher-training/pull/62/commits/ffc7913b522c9a0e9fb901c0bd65b6d9ebf734ee

This took ages to debug :-(

app/form_objects/result_filters/location_filter_form.rb
gpeng: I was just thinking of not accessing the external dependency directly. Happy to leave it as is for now and move it when we use it somewhere else.

app/form_objects/result_filters/location_filter_form.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def initialize(params)
+      @params = params
+      @errors = nil
+      @valid = nil
+      validate
+    end
+
+    def valid?
+      @valid
```
gpeng: I think we'd usually call `validate` in here and then return `errors.empty?`. It would be good to keep this as `ActiveModel` as possible so as it has a familiar interface. You can include `ActiveModel::Validations` and then do things like `validates :option, :presence` etc. You can also use `ActiveModel::Errors` although that might not help you much here. [(example)](https://api.rubyonrails.org/classes/ActiveModel/Errors.html)

TobyRet: @gpeng - I've made the change but not sure about including `ActiveModel::Validations` at this stage - think it's potentially overkill?

app/form_objects/result_filters/location_filter_form.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  private
+
+    def validate
+      case selected_option
+      when NO_OPTION
+        @errors = ["Please choose an option"]
+      when LOCATION_OPTION
+        if location_query.nil?
+          @errors = ["Postcode, town, or city", "Please enter a postcode, city or town in England"]
```
TobyRet: I wonder if all these error strings should be in locales?

TobyRet:
> Looks good. We should probably move all of the other error messages into the locale if we're going to do that here. Not in this PR obvs.

Yep will add a ticket to the backlog board

PR #63, opened 2020-01-29
Use qa api in development mode

timabell:

Use qa api in development mode

So that it works out of the box without having to set up the API too.

If you are actually working on the API then you can override this
setting.

### Guidance to review

:ship:


defong:
Is there something in the readme that suggest that `manage_backend` is needed...

PR #64, opened 2020-01-29
[2341] Add no-js search and filter page for provider filtering

faissalMT:
### Context
As of yet users cannot filter by provider as they can on Search and Compare, in order to replace Search and Compare this functionality should be implemented.
https://www.find-postgraduate-teacher-training.service.gov.uk/?l=3

Implement a search without Javascript and a filter/search results page for searching by provider.

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -8,6 +8,8 @@ module ResultFilters
       if(!params[:l])
         flash[:error] = "Please choose an option"
         redirect_to location_path(params_without_unsupported_location_params)
+      elsif params[:l].to_i == 3
+        redirect_to provider_path(query: params[:query])
       else
         redirect_to results_path(params_without_unsupported_location_params)
       end
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -8,14 +8,18 @@ module ResultFilters
       if(!params[:l])
         flash[:error] = "Please choose an option"
         redirect_to location_path(params_without_unsupported_location_params)
-      elsif params[:l].to_i == 3
-        redirect_to provider_path(query: params[:query])
+      elsif params[:l] == "3"
+        redirect_to provider_path(params_for_provider_search)
       else
         redirect_to results_path(params_without_unsupported_location_params)
       end
     end
 
   private
+
+    def params_for_provider_search
+      filter_params.except(:lat, :lng, :rad, :loc, :lq)
+    end
 
     def params_without_unsupported_location_params
       filter_params.except(:lat, :lng, :rad, :query, :loc, :lq)
--- a/app/controllers/result_filters/provider_controller.rb
+++ b/app/controllers/result_filters/provider_controller.rb
@@ -23,7 +23,10 @@ module ResultFilters
     end
 
     def redirect_back
-      redirect_to location_path(filter_params.except(:query))
+      redirect_params = filter_params
+      redirect_params = redirect_params.except(:query) if params[:query].blank?
+
+      redirect_to location_path(redirect_params)
     end
   end
 end
--- a/app/controllers/result_filters/provider_controller.rb
+++ b/app/controllers/result_filters/provider_controller.rb
@@ -8,10 +8,10 @@ module ResultFilters
         .where(recruitment_cycle_year: "2020")
         .with_params(search: params[:query])
         .all
-    end
 
-    def create
-      redirect_to results_path(filter_params)
+      if @provider_suggestions.count == 1
+        redirect_to results_path(filter_params.merge(query: @provider_suggestions.first.provider_name))
+      end
     end
   end
 end
--- a/app/controllers/result_filters/provider_controller.rb
+++ b/app/controllers/result_filters/provider_controller.rb
@@ -3,6 +3,11 @@ module ResultFilters
     include FilterParameters
 
     def new
+      if params[:query].blank?
+        flash[:error] = "Training provider"
+        return redirect_back
+      end
+
       @provider_suggestions = Provider
         .select(:provider_code, :provider_name)
         .where(recruitment_cycle_year: "2020")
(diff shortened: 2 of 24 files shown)


TobyRet:
Looks like the provider option is expanded by default?
Screenshot 2020-02-03 at 11 56 01

app/controllers/result_filters/location_controller.rb
```diff
@@ -8,13 +8,19 @@ def create
...
+      elsif params[:l] == "3"
+        redirect_to provider_path(params_for_provider_search)
       else
         redirect_to results_path(params_without_unsupported_location_params)
       end
     end
 
   private
 
+    def params_for_provider_search
```
TobyRet: FOR INFO: This is removed in https://trello.com/c/qeSWGWrJ/2783-l-location-filter-add-filter-page-by-locatiion. Should disappear once my or your branch is merged ...

TobyRet:
[[M] location filter: Add filter page - by provider code (Sans JS)](https://trello.com/c/eYlmrBEe/2341-m-location-filter-add-filter-page-by-provider-code-sans-js)

app/controllers/result_filters/provider_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  class ProviderController < ApplicationController
+    include FilterParameters
+
+    def new
+      if params[:query].blank?
+        flash[:error] = "Training provider"
+        return redirect_back
+      end
+
+      @provider_suggestions = Provider
```
TobyRet: Worth moving to a scope?

spec/features/result_filters/location_spec.rb
```diff
@@ -34,6 +36,54 @@
         },
       )
     end
+
+    context "selecting by provider" do
```
TobyRet: Should there be a spec for provider option validation?

spec/features/result_filters/location_spec.rb
```diff
@@ -146,15 +196,14 @@
...
-
-      filter_page.load
+      #Site prism does not correctly handle array arguments
+      visit location_path(test: [1, 2])
       filter_page.across_england.click
       filter_page.find_courses.click
 
       expect(results_page).to be_displayed
 
+
```
TobyRet: Extra line break

spec/site_prism/page_objects/page/result_filters/location.rb
```diff
@@ -6,7 +6,12 @@ class Location < SitePrism::Page
 
         element :back_link, '[data-qa="page-back"]'
         element :error, '[data-qa="error"]'
+        element :error_text, '[data-qa="error__text"]'
+        element :error_heading, '[data-qa="error__heading"]'
         element :across_england, '[data-qa="across_england"]'
+        element :by_provider, '[data-qa="by_provider"]'
```
TobyRet: Think we should be following BEM naming convention here

spec/site_prism/page_objects/page/result_filters/location.rb
```diff
@@ -6,7 +6,12 @@ class Location < SitePrism::Page
...
+        element :provider_error, '[data-qa="provider_error"]'
+        element :provider_search, '[data-qa="provider_search"]'
```
TobyRet: Think we should be following BEM naming convention here

app/controllers/result_filters/location_controller.rb
faissalMT: What's being done in its place?

app/controllers/result_filters/provider_controller.rb
faissalMT: I don't believe JSONApiClient supports scopes, and as it stands this is only used here

app/controllers/result_filters/provider_controller.rb
```diff
@@ -0,0 +1,29 @@
+module ResultFilters
+  class ProviderController < ApplicationController
+    include FilterParameters
+
+    def new
+      if params[:query].blank?
+        flash[:error] = "Training provider"
```
misaka: The error at the top of the page looks like this:

image

Even though the error above the input box looks correct:

image

defong: Copied csharp feature

app/controllers/result_filters/provider_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        return redirect_back
+      end
+
+      @provider_suggestions = Provider
+        .select(:provider_code, :provider_name)
+        .where(recruitment_cycle_year: "2020")
```
misaka: Not liking this being hard-coded here. We should at least use `Settings.current_cycle`. There may also be an argument for letting the API decide what the current cycle is, which would mean not using `where(recruitment_cycle_year: ...)` at all here, but that should be a decision taken across the app.

misaka: Also, this sounds like a candidate for a "scope", or since JsonApiClient doesn't support those directly, could it be a method: `Provider.in_current_cycle` (or similar)?

app/controllers/result_filters/provider_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        .with_params(search: params[:query])
+        .all
+
+      if @provider_suggestions.count.zero?
+        flash[:error] = "Training provider"
```
misaka: Similar issue (?) as above about the error message.

defong:
validation issues

when validation fails
- [ ] back to search results link changes 
- [ ] previously entered value is lost

app/controllers/result_filters/provider_controller.rb
faissalMT: Just double checked this with Misha and it seems to be correct

misaka: Just having a chat with @aliuk2012 and I'm not so sure now @faissalMT, sorry. I'll let Al add his $0.02.

aliuk2012: Ideally, The name in error summary (top part of the page) and the error message near the invalid form field should be identical.

If I were a user I would wonder whether there were two errors and where to correct the second one. 

As a side note note I also spotted that there is a missing red left border across all the radio buttons like this:
image

Perhaps something we could fix that in another PR.


For extra reference: https://design-system.service.gov.uk/components/error-message/

aliuk2012: 🤦‍♂ more noticeable, the input itself is not highlighted. I didn't spot that until I revisited this comment

faissalMT: Since this is present in Find I propose putting it in a seperate PR

app/controllers/result_filters/location_controller.rb
TobyRet: The params are now allowed, that's all

PR #67, opened 2020-01-30
2885-add an app setting

rizzkhan1:
### Context
Add google geocode api key to app settings

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -132,7 +132,7 @@
         "description": "API Key for Google Maps"
       }
     },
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -131,6 +131,12 @@
       "metadata": {
         "description": "API Key for Google Maps"
       }
+    },
+    "settingsGoogleGeoCodeAPIKey": {
```
gpeng: `GeoCode` -> `Geocode` and on line 278

PR #76, opened 2020-02-06
[2902] - Add validation error legend

TobyRet:
### Context

Red legend was not displaying on left hand side when a validation error occurred.
Now it does!

red_legend

### DIFF
--- a/app/views/result_filters/location/new.html.erb
+++ b/app/views/result_filters/location/new.html.erb
@@ -40,11 +40,11 @@
             <div
               class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? %>"
               id="location-conditional">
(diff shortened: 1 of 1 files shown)


app/views/result_filters/location/new.html.erb
```diff
@@ -59,19 +59,19 @@
...
+                                    ["10 miles", "10"],
+                                    ["20 miles", "20"],
+                                    ["50 miles", "50"],
+                                    ["100 miles", "100"],
+                                  ],
+                                  [request.params[:rad] || "20"]
+                                ),
+                                {include_blank: false},
+                                data: {qa: 'search-radius'},
+                                class: 'govuk-select govuk-!-width-one-half'
```
aliuk2012: Do we need so much whitespace?

PR #77, opened 2020-02-06
[2779] Add the subjects filter display to the results page

AgaDufrat:
### Context
Displays the filter state as it exists on the [C# results page](https://find-postgraduate-teacher-training.education.gov.uk/results).

* Subjects are listed in alphabetical order
alphabetical

* "Further education" is treated as just another subject.
* When more than 4 subjects are selected, the first 4 (alphabetically) are listed an the rest are summarised as "and N more..."
more

* When SEND is selected the UI displays "Only SEND courses" at the top of the list of subjects, and it does not count towards the list of 4 subjects and the "and N more..." rule.
send

* Defaults to all subjects, without "Only SEND courses".
defaults

### Guidance to review
Compare with [C# version](https://find-postgraduate-teacher-training.education.gov.uk/results?l=3&query=Essex%20Teacher%20Training&qualifications=QtsOnly,PgdePgceWithQts,Other&fulltime=False&parttime=False&hasvacancies=True&senCourses=False).

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -34,7 +34,7 @@ class ResultsController < ApplicationController
       if params["subjects"].present?
         SubjectArea.includes(:subjects).all
           .map(&:subjects).flatten
-          .select { |subject| rails_to_csharp_subject_id(id: subject.id).in?(params["subjects"].split(",").map(&:to_s)) }
+          .select { |subject| rails_to_csharp_subject_id(id: subject.id).in?(subjects_params_array) }
           .sort_by(&:subject_name)[0..3]
       else
         SubjectArea.includes(:subjects).all
@@ -56,5 +56,9 @@ private
     redirect_uri.query = query_string.gsub("%2C", ",") if query_string.present?
 
     redirect_to redirect_uri.to_s
+  end
+
+  def subjects_params_array
+    params["subjects"].split(",")
   end
 end
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -1,4 +1,5 @@
 class ResultsController < ApplicationController
+  include CsharpRailsSubjectConversionHelper
   before_action :redirect_to_c_sharp
 
   def index
@@ -29,6 +30,17 @@ class ResultsController < ApplicationController
           address4
           postcode
         ]).all
+    @subjects =
+      if params["subjects"].present?
+        SubjectArea.includes(:subjects).all
+          .map(&:subjects).flatten
+          .select { |subject| rails_to_csharp_subject_id(id: subject.id).in?(params["subjects"].split(",").map(&:to_s)) }
+          .sort_by(&:subject_name)[0..3]
+      else
+        SubjectArea.includes(:subjects).all
+          .map(&:subjects).flatten
+          .sort_by(&:subject_name)[0..3]
+      end
   end
 
 private
--- a/app/helpers/csharp_rails_subject_conversion_helper.rb
+++ b/app/helpers/csharp_rails_subject_conversion_helper.rb
@@ -35,7 +35,9 @@ module CsharpRailsSubjectConversionHelper
   end
 
   def csharp_rails_conversion_table
-    [{ csharp_id: "49", rails_id: "42", name: "Modern languages (other)" },
+    [{ csharp_id: nil, rails_id: "27", name: "Philosophy" },
+     { csharp_id: nil, rails_id: "33", name: "Modern Languages" },
+     { csharp_id: "49", rails_id: "42", name: "Modern languages (other)" },
      { csharp_id: "14", rails_id: "43", name: "Further education" },
      { csharp_id: "44", rails_id: "41", name: "Spanish" },
      { csharp_id: "41", rails_id: "40", name: "Russian" },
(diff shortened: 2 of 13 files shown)


AgaDufrat:
[[S] subjects filter: Show filter on results page](https://trello.com/c/62M0SvUT/2779-s-subjects-filter-show-filter-on-results-page)

app/views/results/_subjects.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <h2 class="govuk-heading-s filter-form__title">
+    Subjects<span class="govuk-visually-hidden">:</span>
+  </h2>
+  <ul class="filter-form__value--list">
+    <% if @results_view.send_courses? %>
+      <li data-qa="send_courses">Only SEND courses</li>
+    <% end %>
+
+    <% @subjects.each do |subject| %>
+        <li data-qa="subjects"><%= subject.subject_name %></li>
```
aliuk2012: Very minor, theres a little too much indent here

app/controllers/results_controller.rb
```diff
@@ -29,6 +30,17 @@ def index
...
+      if params["subjects"].present?
+        SubjectArea.includes(:subjects).all
+          .map(&:subjects).flatten
+          .select { |subject| rails_to_csharp_subject_id(id: subject.id).in?(params["subjects"].split(",").map(&:to_s)) }
+          .sort_by(&:subject_name)[0..3]
+      else
+        SubjectArea.includes(:subjects).all
+          .map(&:subjects).flatten
+          .sort_by(&:subject_name)[0..3]
+      end
```
aliuk2012: I may have missed something but I didn't see any tests for this. I was thinking specifically testing that the results are sorting by subject name

AgaDufrat: I didn't add a test because we don't seem to be testing the controllers https://github.com/DFE-Digital/find-teacher-training/tree/master/spec/controllers.
This is tested in the feature test. I should however also test the ordering of the "all subjects" since this is coming from a different part of code.

aliuk2012: Ah we aren't testing the controllers, I thought I had missed something.
Thanks for adding the tests 👍

app/helpers/csharp_rails_subject_conversion_helper.rb
```diff
@@ -35,7 +35,9 @@ def rails_to_csharp_subject_id(id:)
   end
 
   def csharp_rails_conversion_table
-    [{ csharp_id: "49", rails_id: "42", name: "Modern languages (other)" },
+    [{ csharp_id: nil, rails_id: "27", name: "Philosophy" },
+     { csharp_id: nil, rails_id: "33", name: "Modern Languages" },
```
defong: this `Modern Languages` is not needed

app/controllers/results_controller.rb
```diff
@@ -29,6 +30,17 @@ def index
           address4
           postcode
         ]).all
+    @subjects =
+      if params["subjects"].present?
+        SubjectArea.includes(:subjects).all
+          .map(&:subjects).flatten
+          .select { |subject| rails_to_csharp_subject_id(id: subject.id).in?(params["subjects"].split(",").map(&:to_s)) }
```
faissalMT: This line is quite verbose and might benefit from being broken out into a seperate method

app/controllers/results_controller.rb
```diff
@@ -29,6 +30,17 @@ def index
...
+          .sort_by(&:subject_name)[0..3]
```
faissalMT: Why limit to only 3 items? C# Find doesn't seem to be doing this

app/helpers/csharp_rails_subject_conversion_helper.rb
faissalMT: Removing these causes errors because the user can select rails subjects that don't exist on C#. Philosophy is a subject that appears only on Rails but not on C# so maybe we should also prevent the checkbox for it from showing?

misaka:
We've created tickets for the issues surrounding the unused subject "Philosophy" and the mysterious +1 counting issue with on the results page.

https://trello.com/c/ad7EdBHI/2931-remove-the-philosophy-subject-from-filter-page
https://trello.com/c/IYlRUBbK/2933-dont-count-philosophy-and-modern-languages-subjects-in-filter-box

app/helpers/csharp_rails_subject_conversion_helper.rb
AgaDufrat: I checked that the other additional subject that appears is `Modern Languages`

AgaDufrat: I checked that second unused subject is `Modern Languages`

app/controllers/results_controller.rb
```diff
@@ -50,11 +50,14 @@ def redirect_to_c_sharp
...
-
     redirect_uri = URI(base_url)
     redirect_uri.path = "/results"
     redirect_uri.query = query_string.gsub("%2C", ",") if query_string.present?
 
     redirect_to redirect_uri.to_s
   end
+
+  def subjects_params_array
+    params["subjects"].split(",").map(&:to_s)
```
faissalMT:
```suggestion
    params["subjects"].split(",")
```
`.split()` only ever returns an array of strings

PR #81, opened 2020-02-10
[2895] Refactor accordion and checkbox into components

faissalMT:
### Context
We were under the impression that the accordions were expanding due to `aria-expanded=` being set but this was incorrect.  

Set the class `govuk-accordion__section--expanded` when a subject area is selected.

### Guidance to review  
Example behaviour from C#:
https://www.find-postgraduate-teacher-training.service.gov.uk/results/filter/subject?l=2&subjects=0,1,2&qualifications=QtsOnly,PgdePgceWithQts,Other&fulltime=False&parttime=False&hasvacancies=True&senCourses=False

### DIFF
--- a/app/helpers/result_filters/subject_helper.rb
+++ b/app/helpers/result_filters/subject_helper.rb
@@ -13,7 +13,7 @@ module ResultFilters
     def subject_area_is_selected?(subject_area:)
       return false if params["subjects"].nil?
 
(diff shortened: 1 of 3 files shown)


faissalMT:
[All subject accordians expanding together](https://trello.com/c/IIooxhgH/2895-all-subject-accordians-expanding-together)

PR #82, opened 2020-02-10
2896 fix focus colour for pagination link

aliuk2012:
### Context

Part of https://trello.com/c/hfgbMP9N/2896-2d-accessibility-audit-followup-find

While looking at [accessibility ticket](https://trello.com/c/hfgbMP9N) I noticed that Rails version of the pagination links were not styled correctly. Along with the text colour the focus styling was not consistent with the rest of the site and it would also fail WCAG 2.1 colour contrast check.

It will not be fixed in the C# version which currently looks like this:
pagination-c

#### Before
pagination-before

#### After
pagination-after

### Guidance to review
- May need to test this locally so that you can access the rails version on prot 3002 without being redirected to the C# version


aliuk2012:

PR #84, opened 2020-02-11
[2948] Don't include radius when across England search is selected

faissalMT:
### Context
Radius, a parameter used when searching by location, was being included when searching across England

When across England is selected, reject unneeded parameters


app/controllers/result_filters/location_controller.rb
```diff
@@ -9,11 +9,12 @@ def new; end
     def create
       if provider_option_selected?
         redirect_to(provider_path(params_for_provider_search)) && return
+      elsif across_england_option_selected?
```
TobyRet: This should be handled by the LocationFilterForm object

PR #87, opened 2020-02-11
2896 - Add missing id for aria-controls (Accordion component)

aliuk2012:
### Context

Relates to https://trello.com/c/hfgbMP9N/2896-2d-accessibility-audit-followup-find 

A recent accessibility audit completed in December 2019, revealed
that our accordion component was using aria-controls attribut on the
accordion section headings button but no elements on the page had an id
that matched the aria-controls values.

Aria-controls attribute is used by assistive tech to help users understand
what they are interacting with and what elements are responsible for various
actions.

I used govuk-frontend accordion as a reference for this fix. There are
no visual differences but the html should be reviewed to see the changes.

Buttons aria-control value matches the accordion content section for
the specific section it relates too.

Add a unique id to each accordion section content div and ensure that id is used in the sections heading button aria-control attribute

### Guidance to review
No visual difference but check markup

### DIFF
--- a/app/views/result_filters/subject/new.html.erb
+++ b/app/views/result_filters/subject/new.html.erb
@@ -19,17 +19,17 @@
 
 
         <div class="accordion govuk-form-group" data-module="govuk-accordion">
-          <% @subject_areas.each do |subject_area| %>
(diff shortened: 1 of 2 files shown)


aliuk2012:

PR #92, opened 2020-02-12
[2950] Conditionally reveal the content of Location page radios

AgaDufrat:
### Context
When the location was preselected in the query string, the content of the 'By postcode, town or city' radio was not showing.

Selecting 'By postcode, town or city' sets the `l` query param to "1". When this param appears in the query we want to reveal the content of this radio.

| Before       | After          | 
| ------------- |:-------------:| 
| Screenshot 2020-02-12 at 15 03 21| Screenshot 2020-02-12 at 15 03 53|

### Guidance to review
- visit http://localhost:3002/results/filter/location?l=1&lat=51.4980188&lng=-0.1300436&loc=Westminster,+London+SW1P+3BT,+UK&lq=SW1P+3BT&rad=20&qualifications=QtsOnly,PgdePgceWithQts,Other&fulltime=False&parttime=False&hasvacancies=True&senCourses=False
- or just http://localhost:3002/results/filter/location?l=1
- the content for the first radio button should be showing like in [the c# version](https://find-postgraduate-teacher-training.education.gov.uk/?l=1)

### DIFF
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -34,7 +34,7 @@
               <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
             </div>
             <div
-              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? || params[:l] == "1" %>"
+              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless params[:l] == "1" %>"
               id="location-conditional" data-qa="location-conditional">
               <div class="govuk-form-group <%= "govuk-form-group--error" if location_error? %>">
                 <%= form.label :lq, "Postcode, town or city", class: "govuk-label" %>
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -34,8 +34,8 @@
               <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
             </div>
             <div
-              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? %>"
-              id="location-conditional">
+              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? || params[:l] == "1" %>"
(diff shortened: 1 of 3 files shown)


AgaDufrat:
[Rails Find app doesn't expand location filter selection when changing query](https://trello.com/c/hh2JmNqP/2950-rails-find-app-doesnt-expand-location-filter-selection-when-changing-query)

aliuk2012:
Could you fix the `By school` radio as part of this PR?

for example, visit http://localhost:3002/results/filter/location?l=3

app/views/result_filters/location/_form.html.erb
```diff
@@ -34,8 +34,8 @@
               <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
             </div>
             <div
-              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? %>"
-              id="location-conditional">
+              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? || params[:l] == "1" %>"
```
aliuk2012: I was wondering whether we even need `location_error?` we should show the conditional fields regardless of whether there is an error or not. As long as the user has selected the parent radio button in this case "By postcode.." then the conditional fields should be displayed

AgaDufrat: Good point. 

@SamJamCul and I talked about the `By school` radio, which works correctly because it's checking `params[:query]` but I changed it as you suggested for consistency. 

Ready for another look 👀

spec/features/result_filters/location_spec.rb
```diff
@@ -148,10 +148,22 @@
...
+      expect(filter_page.by_postcode_town_or_city.checked?).to eq(true)
+      expect(filter_page.location_conditional).not_to match_selector(".govuk-radios__conditional--hidden")
+    end
+
     it "Preselects across england" do
       filter_page.load(query: { l: 2 })
       expect(filter_page.across_england.checked?).to eq(true)
     end
+
+    it "Preselects by school, university or other training provider" do
```
aliuk2012:
```suggestion
    it "Preselects by school, university or other training provider and reveals the content" do
```

PR #93, opened 2020-02-12
2896 - Place subject areas inside a fieldset

aliuk2012:
After an accessibility audit in December 2019, It was suggested
that each subject area should be placed inside its own fieldset
to help screen reader users understand the context of each
subject area options. It was also suggested that we remove the
main fieldset/legend.

### Guidance to review
No visual changes but you should be able to see the fieldset wrapped around the each group of subject are checkboxes


aliuk2012:

PR #94, opened 2020-02-12
[2839] - live location lookup

TobyRet:
### Context
When a user searches by 'postcode, town or city' on the location filter page, location suggestions are now displayed.

location

### Guidance to review
I expect _DependaSau_ will break this

### DIFF
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -1,0 +1,29 @@
+class LocationSuggestion
+  include HTTParty
+
+  base_uri "https://maps.googleapis.com"
+
+  class << self
+    def suggest(input)
+      query = build_query(input)
+
+      response = get("/maps/api/place/autocomplete/json?#{query.to_query}")
+
+      if response.success?
+        JSON.parse(response.body)["predictions"].map { |p| p["description"] }.take(5)
+      end
+    end
+
+  private
+
+    def build_query(input)
+      {
+          key: Settings.gcp_api_key,
+          language: "en",
+          input: input,
+          components: "country:uk",
+          types: "geocode",
+      }
+    end
+  end
+end
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -1,13 +1,13 @@
 class LocationSuggestion
   include HTTParty
 
-  base_uri "https://maps.googleapis.com"
+  base_uri Settings.google.places_api_host
 
   class << self
     def suggest(input)
       query = build_query(input)
 
-      response = get("/maps/api/place/autocomplete/json?#{query.to_query}")
+      response = get("#{Settings.google.places_api_path}?#{query.to_query}")
 
       if response.success?
         JSON.parse(response.body)["predictions"].map { |p| p["description"] }.take(5)
@@ -18,7 +18,7 @@ class LocationSuggestion
 
     def build_query(input)
       {
-          key: Settings.gcp_api_key,
+          key: Settings.google.gcp_api_key,
           language: "en",
           input: input,
           components: "country:uk",
--- a/app/webpacker/scripts/autocomplete.js
+++ b/app/webpacker/scripts/autocomplete.js
@@ -27,32 +27,32 @@ export const request = endpoint => {
   };
 };
 
-const initAutocomplete = ($el, $input) => {
-  const input = document.getElementById($input);
-  const el = document.getElementById($el);
+const initAutocomplete = ({element, input, path}) => {
+  const $input = document.getElementById(input);
+  const $el = document.getElementById(element);
   const inputValueTemplate = result => (typeof result === "string" ? result : result && result.name);
   const suggestionTemplate = result =>
     typeof result === "string" ? result : result && `${result.name} (${result.code})`;
 
   try {
-    if(input) {
+    if($input) {
       accessibleAutocomplete({
-        element: el,
-        id: input.id,
+        element: $el,
+        id: $input.id,
         showNoOptionsFound: true,
-        name: input.name,
-        defaultValue: input.value,
+        name: $input.name,
+        defaultValue: $input.value,
         minLength: 3,
-        source: request(input.getAttribute("data-url")),
+        source: request(path),
         templates: {
           inputValue: inputValueTemplate,
           suggestion: suggestionTemplate
         },
-        onConfirm: option => (input.value = option ? option.code : ""),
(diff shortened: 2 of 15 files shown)


app/models/location_suggestion.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  base_uri "https://maps.googleapis.com"
+
+  class << self
+    def suggest(input)
+      query = build_query(input)
+
+      response = get("/maps/api/place/autocomplete/json?#{query.to_query}")
+
+      if response.success?
```
TobyRet: More / better error handling required?

TobyRet:
[[M] location filter: Live address lookup](https://trello.com/c/5j7Vwnpw/2839-m-location-filter-live-address-lookup)

app/services/location_suggestion.rb
```diff
@@ -0,0 +1,29 @@
+class LocationSuggestion
+  include HTTParty
+
+  base_uri "https://maps.googleapis.com"
```
defong: Maybe stick these into settings?

app/services/location_suggestion.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  class << self
+    def suggest(input)
+      query = build_query(input)
+
+      response = get("/maps/api/place/autocomplete/json?#{query.to_query}")
```
defong: Settings as well maybe?

app/webpacker/scripts/autocomplete.js
```diff
@@ -32,7 +32,7 @@ const initAutocomplete = ($el, $input) => {
   const el = document.getElementById($el);
   const inputValueTemplate = result => (typeof result === "string" ? result : result && result.name);
   const suggestionTemplate = result =>
-    typeof result === "string" ? result : result && `${result.name} (${result.providerCode})`;
+    typeof result === "string" ? result : result && `${result.name} (${result.code})`;
```
defong: Does this change break the provide code?

app/views/result_filters/location/new.html.erb
```diff
@@ -50,12 +50,14 @@
                 <%=
                   form.text_field(
                     :lq,
+                    id: 'location',
                     value: "#{request.params[:lq]}",
                     class: "govuk-input",
-                    data: {qa: "location-query"},
+                    data: {qa: "location-query", url: "/location-suggestions"},
```
defong: Is this url same as location_suggestions_path (railsy one)

app/webpacker/scripts/autocomplete.js
```diff
@@ -32,7 +32,7 @@ const initAutocomplete = ($el, $input) => {
   const el = document.getElementById($el);
   const inputValueTemplate = result => (typeof result === "string" ? result : result && result.name);
   const suggestionTemplate = result =>
```
defong: inputValueTemplate and suggestionTemplate functions can be move out of initAutocomplete.

app/views/result_filters/location/new.html.erb
```diff
@@ -38,7 +38,7 @@
               <%= form.label :l_1, "By postcode, town or city", class: "govuk-label govuk-radios__label" %>
             </div>
             <div
-              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless location_error? %>"
+              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless request.params[:lq] || location_error? %>"
```
aliuk2012: Do we even need to include `location_error?`? shouldn't we always show the conditional fields if the parent radio button has been set.

aliuk2012: Is this the same change as whats coming in #92 ?

aliuk2012:
```suggestion
              class="govuk-radios__conditional <%= "govuk-radios__conditional--hidden" unless request.params[:lq] %>"
```

app/services/location_suggestion.rb
TobyRet: Woops. Will do

app/webpacker/scripts/autocomplete.js
TobyRet: It does the opposite. It unbreaks it 😄 ... The code highlighted in red is C# . I'll add as a separate commit

app/views/result_filters/location/new.html.erb
TobyRet: Yep. You rails ninja

app/webpacker/scripts/autocomplete.js
TobyRet: I don't see why you would want to do this? They should be scoped to the function as they are not used elsewhere

app/views/result_filters/location/new.html.erb
TobyRet: Ah yes true

app/webpacker/scripts/autocomplete.js
aliuk2012: I agree with @TobyRet, if its not being used else where then it should be scoped to the function.

app/views/result_filters/location/new.html.erb
```diff
@@ -121,7 +123,7 @@
                                     id: "provider",
                                     value: params[:query],
                                     class: "govuk-input",
-                                    data: {qa: "provider-search"} %>
+                                    data: {qa: "provider-search", url: "/provider-suggestions"} %>
```
aliuk2012: Another rails path?

app/models/location_suggestion.rb
TobyRet: Any other exceptions will automatically go to Sentry ...

app/views/result_filters/location/new.html.erb
TobyRet: Yep got that

app/webpacker/scripts/autocomplete.js
```diff
@@ -43,7 +43,7 @@ const initAutocomplete = ($el, $input) => {
         name: input.name,
         defaultValue: input.value,
         minLength: 3,
-        source: request("/provider-suggestions"),
+        source: request(input.getAttribute("data-url")),
```
aliuk2012: could this be open to xss? Just thinking if the attribute is replaced with something malicious then the lookup could be use as a keylogger.

TobyRet: Path is now being passed as param to initAutocomplete function

PR #96, opened 2020-02-13
2896 - Explain abbreviations for the first time.

aliuk2012:
An issue raised in a recent accessibility audit.

There are abbreviated words present on the page (PGCE with QTS) that have
not been explained in their first instance.


aliuk2012:

aliuk2012:
@defong  I passed your comment onto Paul G and his response was that it is complicated and something that was discussed in the early days of the service but we decided to go with the current case as we have in this PR.


Thanks for raising this.

PR #98, opened 2020-02-13
[2966] change course model link to provider code

SamJamCul:
Co-authored-by: Faissal Bensefia <faissal@madetech.com>

### Context

[This](https://trello.com/c/tvPnfnbr/2966-find-breaks-when-making-api-call-using-provider-code) Trello card has it

Change the course model to use the shallow_path flag for course codes


SamJamCul:
[Find: breaks when making api call using provider code](https://trello.com/c/tvPnfnbr/2966-find-breaks-when-making-api-call-using-provider-code)

PR #99, opened 2020-02-13
[2839] - Update Azure template for accessing Google geocode api key

TobyRet:
### Context

This is because the rails settings were updated in https://github.com/DFE-Digital/find-teacher-training/pull/94 .


TobyRet:
[[M] location filter: Live address lookup](https://trello.com/c/5j7Vwnpw/2839-m-location-filter-live-address-lookup)

PR #102, opened 2020-02-14
[2960] Add subject filter validation

AgaDufrat:
### Context
There is no validation on the subject filter, that exists in the [C# version](https://find-postgraduate-teacher-training.education.gov.uk/start/subject?l=2).

Add validation and display an error to the user. Error messages is linked to the first checkbox.

subject_validation

### Guidance to review
- visit http://localhost:3002/start/subject?l=2
- click Continue
- _error should appear_
- click Please choose at least one subject
- _it should be linked to the first checkbox ([like in this example](https://design-system.service.gov.uk/components/error-summary/linking-checkboxes-radios/index.html)) and the first accordion should be expanded_

**Review questions:**
- Do we need to be testing the error text "Please choose at least one subject" ?
- Should `no_subject_selected_error?` return `false` instead of `nil`?

Note: I will wait for Graham & Toby to merge their changes first and rebase with master (and change the tests in line with that they did).

### DIFF
--- a/app/controllers/result_filters/subject_controller.rb
+++ b/app/controllers/result_filters/subject_controller.rb
@@ -14,7 +14,12 @@ module ResultFilters
     def start; end
 
     def create
-      redirect_to results_path(filter_params.merge(subjects: convert_rails_subject_id_params_to_csharp))
+      if params[:subjects].blank? && (params[:senCourses].blank? || params[:senCourses] == "false")
+        flash[:error] = [I18n.t("subject_filter.errors.no_option")]
+        redirect_to subject_path(filter_params)
+      else
+        redirect_to results_path(filter_params.merge(subjects: convert_rails_subject_id_params_to_csharp))
+      end
     end
 
   private
--- a/spec/helpers/result_filters/subject_helper_spec.rb
+++ b/spec/helpers/result_filters/subject_helper_spec.rb
@@ -12,4 +12,16 @@ feature "ResultFilters::SubjectHelper", type: :helper do
       expect(helper.subject_is_selected?(id: "4")).to eq(false)
     end
   end
+
+  describe "#no_subject_selected_error?" do
+    it "returns true if no subject is selected" do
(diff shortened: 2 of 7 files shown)


app/controllers/result_filters/subject_controller.rb
```diff
@@ -14,7 +14,12 @@ def new; end
     def start; end
 
     def create
-      redirect_to results_path(filter_params.merge(subjects: convert_rails_subject_id_params_to_csharp))
+      if params[:subjects].blank?
```
defong: my bad needs to check if `SEND` 

https://github.com/DFE-Digital/search-and-compare-ui/blob/9130e9538fd2848f87bf4f21da09d44a8227ebd0/src/Controllers/FilterController.cs#L71-L75

AgaDufrat: Good point, I forgot SEND is a separate thing

spec/helpers/result_filters/subject_helper_spec.rb
```diff
@@ -12,4 +12,16 @@
...
+
+  describe "#no_subject_selected_error?" do
+    it "returns true if no subject is selected" do
+      allow(controller).to receive(:flash).and_return(error: I18n.t("subject_filter.errors.no_option"))
+
+      expect(helper.no_subject_selected_error?).to eq(true)
+    end
+
+    it "returns nil if there is no subject selected" do
+      expect(helper.no_subject_selected_error?).to eq(nil)
```
defong: > Should no_subject_selected_error? return false instead of nil?

Very good question...

I say it should just really be `true|false` as there isn't much to it and there is nothing can be operated in the whatever object that it can returned, in either case, I'm not fussed.

Nice that you made a test of it to illustrate it. :+1: 


@misaka your call or anyone else for final verdict/team vote.
image

dankmitchell: The method is a question so I would expect true or false

defong:
[Add subject filter validation](https://trello.com/c/qvH3GwiP/2960-add-subject-filter-validation)

PR #106, opened 2020-02-17
[2760] Sort results by provider name (new version)

faissalMT:
### Context
We want to be able to sort the course list by provider name, in order to do this Find needs to request the ordered version of the course list. Additionally you should be able to switch between the descending and ascending order.
https://github.com/DFE-Digital/teacher-training-api/pull/1172

Add a sort form for subjects, with automatic submission when Javascript is enabled and automatic selection.

### Guidance to review
The 'sorted by' portion of this:
https://www.find-postgraduate-teacher-training.service.gov.uk/results?l=2&subjects=53&qualifications=QtsOnly,PgdePgceWithQts,Other&fulltime=False&parttime=False&hasvacancies=True&senCourses=False


dankmitchell:
Heroku review app works - https://find-teacher-2760-sort-0rxzi9y.herokuapp.com/results

faissalMT:
Card for doing the backend for that @defong https://trello.com/c/L3OkEzTk/2993-allow-sorting-courses-by-course-name-as-well-as-provider-name

app/controllers/results_controller.rb
```diff
@@ -4,6 +4,7 @@ class ResultsController < ApplicationController
 
   def index
     @results_view = ResultsView.new(query_parameters: request.query_parameters)
+
     @courses = Course
```
gpeng: This has all been moved into `ResultsView` now. The course bit in #109 and the subjects in #115 which is still in review.

app/controllers/results_controller.rb
```diff
@@ -44,6 +48,12 @@ def index
 
 private
 
+  def get_order
```
gpeng: This will go in `ResultsView` too. I don't think we need the `get_` prefix.

app/views/result_filters/funding/new.html.erb
```diff
@@ -15,7 +15,7 @@
         Salaried courses are not eligible for bursaries, scholarships or student finance. There are usually no course fees to pay.
       </p>
       <%= form_with url: funding_path do |form| %>
-        <%= render 'result_filters/hidden_fields', exclude_keys: ["funding"], form: form %>
+        <%= render "shared/hidden_fields", exclude_keys: ["funding"], form: form %>
```
gpeng: This is only shared amongst the `result_filter` views so I'd leave it in that folder. I can live it with it moving though if you disagree 👍

app/views/results/index.html.erb
```diff
@@ -43,6 +43,41 @@
           </p>
         </div>
       </div>
+      <% unless params[:l] == "3" %>
```
gpeng: Can we make this a method in `ResultsView`? Then we can test it more easily and name it so we know what `params[:l] == "3"` means.

app/views/result_filters/funding/new.html.erb
faissalMT: I moved it because I need it on the results page too

gpeng: Ok... fair enough. Should we name the file something different then do you think? "shared/result_querystring_hidden_fields" or something like that. Just thinking 'hidden_fields' doesn't really give you much to go on about what it is.

gpeng: Fair enough. What do you think about renaming the file itself then? 'hidden_fields' doesn't give a lot of description of what the file is. What about `results_querystring_hidden_fields` or something like that?

spec/features/results_spec.rb
```diff
@@ -143,6 +183,76 @@
...
         expect("Biology").to appear_before("Primary")
       end
+
+      context "with parameters" do
+        let(:sort) { "-provider.provider_name,-name" }
+
+        let(:results_page_request) do
+          stub_request(:get, default_url)
+            .with(query: {
+              "filter[vacancies]" => "true",
```
gpeng: We've got `base_parameters` that we could use here. I agree with @danielburnley about maybe extracting that for reuse. I'm just starting on another 'do the filtering' card so I'm happy to address that there if it's a pain to pull out in this PR.

PR #107, opened 2020-02-17
[2970] Return user back to location wizard on validation error

defong:
### Context
Return user back to location wizard on validation error if they started on location wizard and vice versa for location filter page

Fixed validation
After validation users ends up on the root path instead of location path

### Guidance to review
:shipit:


defong:
[[S] Return user back to location wizard on validation error](https://trello.com/c/JeEBmezy/2970-s-return-user-back-to-location-wizard-on-validation-error)

PR #108, opened 2020-02-17
[2954] Return user back to subject wizard on validation error

defong:
After validation users ends up on the subject path vs start subject path

### Context
Return user back to subject wizard on validation error if they started on subject wizard and vice versa for subject filter page

Fixed validation
After validation users ends up on the start subject path instead of subject path

### Guidance to review
:shipit:

### DIFF
--- a/app/controllers/result_filters/subject_controller.rb
+++ b/app/controllers/result_filters/subject_controller.rb
@@ -11,12 +11,19 @@ module ResultFilters
 
     def new; end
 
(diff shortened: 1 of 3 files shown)


defong:
[[S] Return user back to subject wizard on validation error](https://trello.com/c/dHaW3ZPV/2954-s-return-user-back-to-subject-wizard-on-validation-error)

PR #109, opened 2020-02-18
[2979] - add filtering for salary, qualification, vacancies & study type

TobyRet:
### Context

Adds filtering for salary, qualification, vacancies and study type.

When filtering, calls can now be made to V3 API 😸 

### Guidance to review
- This is a big PR
- There is very little logic added to make the filtering work. However there are lots of changes in the specs. As discussed at the last tech brakedance we propose that specs should use a simpler method for stubbing JSON responses from the V3 API - i.e. stub at the network level rather than the JSON API resource layer.
- We have moved a lot of the specs out of the `results_spec` and put them into the filter specs. This is because feature specs should be testing the entire feature end to end, not just a page.

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -105,6 +105,7 @@ class ResultsView
 
                    base_query = base_query.where(funding: "salary") if with_salaries?
                    base_query = base_query.where(vacancies: hasvacancies?)
+                   base_query = base_query.where(study_type: study_type) if study_type.present?
 
                    base_query
                      .page(query_parameters[:page] || 1)
@@ -167,5 +168,11 @@ private
     else
       "14"
     end
+  end
+
+  def study_type
+    return "full_time,part_time" if fulltime? && parttime?
+    return "full_time" if fulltime?
+    return "part_time" if parttime?
   end
 end
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -114,8 +114,8 @@ class ResultsView
                  end
   end
 
-  def total_course_count
-    courses.total_count
+  def course_count
+    courses.count
   end
 
 private
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -104,6 +104,7 @@ class ResultsView
                      .where(recruitment_cycle_year: Settings.current_cycle)
 
                    base_query = base_query.where(funding: "salary") if with_salaries?
+                   base_query = base_query.where(vacancies: hasvacancies?)
 
                    base_query
                      .page(query_parameters[:page] || 1)
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -81,12 +81,12 @@ class ResultsView
 
   def map_image_url
     "#{Settings.google.maps_api_url}\
-    ?key=#{Settings.google.maps_api_key}\
-    &center=#{latitude},#{longitude}\
-    &zoom=#{google_map_zoom}\
-    &size=300x200\
-    &scale=2\
-    &markers=#{latitude},#{longitude}"
+?key=#{Settings.google.maps_api_key}\
+&center=#{latitude},#{longitude}\
+&zoom=#{google_map_zoom}\
+&size=300x200\
+&scale=2\
+&markers=#{latitude},#{longitude}"
   end
 
   def provider
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -106,7 +106,7 @@ class ResultsView
                    base_query = base_query.where(funding: "salary") if with_salaries?
                    base_query = base_query.where(vacancies: hasvacancies?)
                    base_query = base_query.where(study_type: study_type) if study_type.present?
-                   base_query = base_query.where(qualifications: query_parameters["qualifications"]) if query_parameters["qualifications"].present?
+                   base_query = base_query.where(qualifications: qualifications)
 
                    base_query
                      .page(query_parameters[:page] || 1)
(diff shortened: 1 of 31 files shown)


app/views/results/index.html.erb
```diff
@@ -26,7 +26,7 @@
     <div class="govuk-grid-column-two-thirds">
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-one-half">
-          <p class="govuk-body search-results__count"><%= number_with_delimiter(@courses.total_count) %> courses found</p>
+          <p class="govuk-body search-results__count"><%= number_with_delimiter(@results_view.total_course_count) %> courses found</p>
```
defong: Zero courses issues

spec/fixtures/api_responses/courses.json
```diff
@@ -0,0 +1,1065 @@
...
+          "contact_name": "Programme Director",
+          "year_code": "2019",
+          "provider_code": "E69",
+          "provider_type": "scitt",
+          "postcode": "EC3N 3AJ",
+          "website": "http://www.e-qualitas.org.uk",
+          "address1": "6th Floor, 10 Lloyd's Avenue",
+          "address2": "",
+          "address3": "London",
+          "email": "initialteachertraining@e-qualitas.co.uk",
```
defong: we shouldn't using email address that not ours

i fact shouldn't this be dummy data... not want to see this pop up in a google search

danielburnley: ^^ If we're using fixtures they should still be dummy data

spec/features/result_filters/location_spec.rb
```diff
@@ -5,10 +5,35 @@
...
+        "page[page]" => 1,
+        "page[per_page]" => 10,
+    }
+  end
 
   before do
     stub_geocoder
-    stub_results_page_request
+
+    stub_api_v3_resource(
```
gpeng: I've just been having a look through this method. I think it bypasses the normal deserialization of the json and uses tests specific stuff. If that is what's happening I think for feature specs where we're trying to test the full stack we should stub the network response with webmock. That way we are testing the code in the model as well. We could use `stub_api_v3_request` for that I think but it takes a hash so we'd have the overhead of getting an example response in JSON and converting it. That's ok for small bits of data but in these specs the responses are quite big.

spec/fixtures/api_responses/courses.json
gpeng: Sorry.. my fault. I thought the data was anonymised.

TobyRet: @defong @danielburnley 
QA, Staging and Prod data is not anonymised. It's public data - I’m not sure I see what the issue is?

spec/features/result_filters/location_spec.rb
TobyRet: 👍

app/view_objects/results_view.rb
```diff
@@ -141,4 +162,14 @@ def google_map_zoom
       "14"
     end
   end
+
+  def study_type
+    return "full_time,part_time" if fulltime? && parttime?
```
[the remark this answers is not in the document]
TobyRet: Sorry not sure I understand @defong

app/views/results/index.html.erb
```diff
@@ -1,9 +1,9 @@
-<%= content_for :page_title, "#{number_with_delimiter(@courses.total_count)} courses" %>
+<%= content_for :page_title, "#{number_with_delimiter(@results_view.total_course_count)} courses" %>
```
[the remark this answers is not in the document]
TobyRet: @defong - this has not been implemented yet. Should be done as part of https://trello.com/c/aHtGYeOY/2838-add-suggested-searches-to-location-searching

app/views/results/index.html.erb
TobyRet: Yep count needs to be fixed 👍

app/views/results/index.html.erb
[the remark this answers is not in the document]
defong: Page title... should be in another card altogether...

app/view_objects/results_view.rb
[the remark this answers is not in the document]
defong: when both `full_time` and `part_time` are `false` than they are defaulted to be both `true`

spec/fixtures/api_responses/courses.json
defong: We only need dummy data for the structure.

Everything else is excessive.

app/views/results/index.html.erb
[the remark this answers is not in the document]
TobyRet: 👍

app/views/results/index.html.erb
TobyRet: https://github.com/DFE-Digital/find-teacher-training/pull/109/commits/712699839ae0ab984a8ad7897565bc97ff4bc0d7

app/view_objects/results_view.rb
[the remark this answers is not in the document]
TobyRet: Ok makes sense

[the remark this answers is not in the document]
TobyRet: https://github.com/DFE-Digital/find-teacher-training/pull/109/commits/084f52e026201957142785b7a0aa9cd90dc6b863

spec/fixtures/api_responses/courses.json
TobyRet: Data has been anonymised

app/view_objects/results_view.rb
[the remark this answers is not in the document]
gpeng: I don't think we need to do this. The Rails API returns all study modes by default so we don't need to pass a filter if we want all of them. We only need the filter if we want to restrict it to just full time or just part time.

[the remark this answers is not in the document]
TobyRet: OK - I can just drop this commit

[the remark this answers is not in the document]
defong: :+1:

PR #111, opened 2020-02-18
2968 - Change how we load in Google Analytics

aliuk2012:
We used to automatically add in Google Analytics (GA), we now 
have to ask users for consent before adding cookies to their devices.
At the same time we should only inject GA if the user consents
otherwise they are downloading more javascript that will not be
used or needed.

This commit removes GTM (we decided to not replace GA with GTM).
GA should still load as before but we will add the cookie policy
control to this to ensure it only gets used if the user consents
to cookies

### DIFF
--- a/app/webpacker/scripts/analytics.spec.js
+++ b/app/webpacker/scripts/analytics.spec.js
@@ -17,19 +17,19 @@ describe("Analytics", () => {
       document.head.appendChild(document.createElement('script'));
     })
   
-    it('adds dynamically adds a script tag', () => {
+    it('dynamically adds a script tag', () => {
       loadAnalytics()
       // GTM will have a src and it will be the only script tag
       expect(document.querySelectorAll('script[src]')).toHaveLength(1)
     })
 
-    it ('takes uses a default tracking ID', () => {
+    it ('uses a default tracking ID', () => {
       loadAnalytics()
       const createGAInstanceWithID = ga.mock.calls[0][1]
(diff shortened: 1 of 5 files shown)


app/webpacker/scripts/analytics.spec.js
```diff
@@ -7,6 +10,45 @@ describe("Analytics", () => {
...
+      expect(document.querySelectorAll('script[src]')).toHaveLength(1)
+    })
+
+    it ('takes uses a default tracking ID', () => {
+      loadAnalytics()
+      const createGAInstanceWithID = ga.mock.calls[0][1]
+      expect(createGAInstanceWithID).toEqual('UA-112932657-1')
+    })
+
+    it ('takes accepts a custom tracking ID', () => {
```
fofr: `takes` or `accepts`

fofr:
Does this PR have the cookie check, or will that come later?

app/webpacker/scripts/analytics.spec.js
aliuk2012: Good spot 😄 👍

aliuk2012:
@fofr  Still to come. I've broken up the PR so that its easier to review instead of a one large PR.


I'll create a label for cookies and label each PR relating to the cookies work

aliuk2012:
[[M] Build and release new cookie banner and cookie policy page.](https://trello.com/c/OnWGvVre/2968-m-build-and-release-new-cookie-banner-and-cookie-policy-page)

PR #112, opened 2020-02-18
2968 Hide cookie banner on cookie page

aliuk2012:
### Context
Users should not be prompted with the cookie banner on the cookie page. They can set their preferences using the new form which will be added to it.

### Guidance to review
Currently the banner is not visible but if you go to the the cookies page you should not see any markup in the html relating to the cookie banner.


aliuk2012:
[[M] Build and release new cookie banner and cookie policy page.](https://trello.com/c/OnWGvVre/2968-m-build-and-release-new-cookie-banner-and-cookie-policy-page)

PR #113, opened 2020-02-18
[2992] fix back link on course show page

misaka:
### Context

Back link on course page is broken.

Fix back link on course page.

### Guidance to review

Tricky to test locally, but hopefully the tests are correct now.

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -2,7 +2,7 @@
 
 <% if permitted_referrer? %>
   <%= content_for(:before_content) do %>
-    <%= govuk_back_link_to(results_path(request.referrer), "Back to search results") %>
+    <%= govuk_back_link_to(request.referrer, "Back to search results") %>
   <% end %>
 <% end %>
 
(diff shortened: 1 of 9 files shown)


app/views/courses/show.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <% if permitted_referrer? %>
   <%= content_for(:before_content) do %>
-    <%= govuk_back_link_to(results_path(request.referrer), "Back to search results") %>
+    <%= govuk_back_link_to(request.referrer, "Back to search results") %>
```
aliuk2012: Is there any chance someone could access this page directly and therefore request.referrer would not be set?

misaka: Yes, but I think this is the way C# did it so we're replicating that? Can't test this in prod now because we've enabled redirects.

misaka: @dankmitchell Do you remember offhand?

dankmitchell: If they access the page directly we don't show the back link. Thats intentional.

fofr: We should show an alternative link, along the lines of "< Start a search for courses" or something. But this matches what's live for now.

The lack of a back link here never sat well with me.

misaka:
[Back link not working in Find](https://trello.com/c/eonJDndw/2992-back-link-not-working-in-find)

PR #115, opened 2020-02-19
[2963] Do the filtering subject

gpeng:
### Context
Actually filter results when the subject filter page is submitted. Part of the work to rebuild Find.

* Use subject codes internally instead of the rails DB ids. The teacher-training-api call requires the codes not the IDs
* Move the subject retrieval for the results page into the results view
* Add feature specs that test the feature
* Remove the filter specs from the results page feature to avoid duplication
* Fix all of the other tests in the subject filter feature. Most of these don't belong in there but it's outside of the scope of this work to move them for now. Probably would have been quicker than getting them to work but hindsight is a wonderful thing. They are mostly either tests that should be done in view specs testing how the template renders data or tests that are part of other features e.g. the start page.

### Guidance to review

The API work for this hasn't been merged as yet. https://github.com/DFE-Digital/teacher-training-api/pull/1186

The existing tests in the subjects feature relied on a set of stubbed out `SubjectArea`-ses. The feature specs stub at the network. I've updated the existing ones to use the network stub. This probably wasn't the best idea as I didn't realise how much I'd have to change them all when I started. They're going to be moved anyway so I've left it as is for now.


gpeng:
[[M] Do the filtering - subject](https://trello.com/c/pqHbUihP/2963-m-do-the-filtering-subject)

spec/fixtures/api_responses/subject_areas.json
```diff
@@ -0,0 +1 @@
...ct_knowledge_enhancement_course_available":true}},{"id":"39","type":"subjects","attributes":{"subject_name":"Mandarin","subject_code":"20","bursary_amount":"26000","early_career_payments":"2000","scholarship":null,"subject_knowledge_enhancement_course_available":true}},{"id":"40","type":"subjects","attributes":{"subject_name":"Russian","subject_code":"21","bursary_amount":"26000","early_career_payments":"2000","scholarship":null,"subject_knowledge_enhancement_course_available":true}},{"id":"41","type":"subjects","attributes":{"subject_name":"Spanish","subject_code":"22","bursary_amount":"26000","early_career_payments":"2000","scholarship":"28000","subject_knowledge_enhancement_course_available":true}},{"id":"42","type":"subjects","attributes":{"subject_name":"Modern languages (other)","subject_code":"24","bursary_amount":"26000","early_career_payments":"2000","scholarship":null,"subject_knowledge_enhancement_course_available":true}},{"id":"43","type":"subjects","attributes":{"subject_name":"Further education","subject_code":"41","bursary_amount":null,"early_career_payments":null,"scholarship":null,"subject_knowledge_enhancement_course_available":null}}],"jsonapi":{"version":"1.0"}}
```
TobyRet: Can your IDE easily format the JSON like so? https://github.com/DFE-Digital/find-teacher-training/blob/master/spec/fixtures/api_responses/courses.json

- Makes a bit more readable but not a deal breaker

gpeng: 🤣 My 'IDE' can barely even open that file. I've tried formatting it but there must be an error in it somewhere when I do as it starts failing loads of tests with a weird seemingly unrelated error message. I'm going to merge and I'll sort that out in a different PR.

PR #117, opened 2020-02-20
[3004] - Return status 400 (bad request) for invalid provider suggestion query

TobyRet:
### Context
The provider autocomplete feature uses a local endpoint ('/provider-suggestions') to query for providers. The autocomplete library is configured to only fire once three characters have been entered in the input field. However because the endpoint is publicly accessible it would be possible to query the endpoint with an invalid query (e.g. less than two characters). 

client_error

The provider suggestion query is now validated before passing through to the Teacher Training API. If invalid a 400 response is returned.

400 error

### DIFF
--- a/spec/requests/provider_suggestions_spec.rb
+++ b/spec/requests/provider_suggestions_spec.rb
@@ -1,0 +1,52 @@
+require "rails_helper"
+
+describe "/provider-suggestions", type: :request do
+  context "when provider suggestion is blank" do
+    it "returns bad request (400)" do
+      get "/provider-suggestions"
+
+      expect(response.status).to eq(400)
+      expect(JSON.parse(response.body)).to eq("error" => "Bad request")
+    end
+  end
+  context "when provider suggestion is less than three characters" do
+    it "returns bad request (400)" do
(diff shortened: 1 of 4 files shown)


spec/requests/provider_suggestions_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+  context "when provider suggestion is less than three characters" do
+    it "returns bad request (400)" do
+      get "/provider-suggestions?query=St"
+
+      expect(response.status).to eq(400)
+      expect(JSON.parse(response.body)).to eq("error" => "Bad request")
+    end
+  end
+  context "when provider suggestion ids valid" do
```
danielburnley: Could we add a test in here to make sure the stubbed request is made? :eyes:

danielburnley: (And the subsequent handling of the response)

TobyRet: 👍

dankmitchell:
As discussed, we need to stop Google crawling this endpoint

PR #119, opened 2020-02-22
[3015] Add XML Sitemap

dankmitchell:
### Context
Sitemap

Add an XML sitemap to replace https://www.find-postgraduate-teacher-training.service.gov.uk/sitemap.txt, not sure why we opted for a TXT version in the c# version when XML is the default.

Note: We need the API to return *all* records and not just the first 10

### Guidance to review
`/sitemap.xml`

Screenshot 2020-02-22 at 18 09 04

### DIFF
--- a/app/controllers/sitemaps_controller.rb
+++ b/app/controllers/sitemaps_controller.rb
@@ -1,0 +1,8 @@
+class SitemapsController < ApplicationController
+  def show
+    @courses = Course
+      .includes(:provider)
+      .where(recruitment_cycle_year: Settings.current_cycle)
+      .all
+  end
+end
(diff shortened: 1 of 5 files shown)


dankmitchell:
[Add sitemap](https://trello.com/c/2TRC1KAO/3015-add-sitemap)

app/controllers/sitemaps_controller.rb
```diff
@@ -0,0 +1,8 @@
+class SitemapsController < ApplicationController
+  def show
+    @courses = Course
```
TobyRet: Probably more efficient to use `pluck` ....

```
Course.includes(:provider).
  where(recruitment_cycle_year: Settings.current_cycle).
  pluck(:course_code, :provider_code)
```

fofr:
@dankmitchell We used text because it was easier at the time, it made a smaller file and didn't need pagination.

PR #124, opened 2020-02-24
[3019] - Display correct course count on results page

TobyRet:
### Context

Use Json Api gem's 'total_count' method rather than 'count'. The latter provides the count of the paginated response (i.e. X per page) rather than the total results count

count

- use of correct method
- added spec


TobyRet:
[[S] Course count on results page only displays 10](https://trello.com/c/tBI9T9QE/3019-s-course-count-on-results-page-only-displays-10)

PR #125, opened 2020-02-24
[3041] - Remove England geocoding restriction

TobyRet:
### Context
- We do not want to filter geocoding results as there may be users who wish to search for courses near England (e.g Scotland, Wales) within the accepted ranges

- Removes England geocoding restriction
- Brings functionality in line with C# legacy application

### DIFF
--- a/app/form_objects/result_filters/location_filter_form.rb
+++ b/app/form_objects/result_filters/location_filter_form.rb
@@ -41,13 +41,12 @@ module ResultFilters
     end
 
     def geocode_params_for(query)
-      results = Geocoder.search(query, components: "country:UK")
-      england_results = results.select { |r| r.state == "England" }.first
-      if england_results
+      results = Geocoder.search(query, components: "country:UK").first
+      if results
         {
-            lat: england_results.latitude,
-            lng: england_results.longitude,
-            loc: england_results.address,
+            lat: results.latitude,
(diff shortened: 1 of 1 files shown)


fofr:
Does this have any effect when searching for place names that can be in England or not in England, eg Boston?

TobyRet:
> Does this have any effect when searching for place names that can be in England or not in England, eg Boston?

No - we still are restricting searches to the UK
See - https://github.com/DFE-Digital/find-teacher-training/blob/745eadeee3e0bf13bad709f196ffaa815d22f552/app/form_objects/result_filters/location_filter_form.rb#L44

dankmitchell:
[Remove England geocoding restriction from location searching](https://trello.com/c/w8TzFv6V/3041-remove-england-geocoding-restriction-from-location-searching)

aliuk2012:
Could we add some tests for this?

TobyRet:
> Could we add some tests for this?

@aliuk2012 - the geocoder gem comes with test stub helpers 
https://github.com/DFE-Digital/find-teacher-training/blob/f7725ac705aeddd698806dfca42973f4334bd7eb/spec/support/geocoder_helper.rb#L1

We don't need a regression test for this change but I think at the time when the England restriction was added another test could have been added as well.

PR #127, opened 2020-02-24
2957 m provider filter do the filtering

gpeng:
### Context

Ongoing work to implement filtering.

* Update some page object references to make them consistent
* Hook up provider filtering
* Move some specs from results that are now covered in the provider feature specs
* Remove some unused page object references

### Guidance to review

Feature specs only cover the non-javascript versions of functionality. I did have a quick go turning javascript specs on but there were some issues with the app startup hanging so to avoid a rabbit hole I've left it for this PR. There is a card to add them https://trello.com/c/30imjCHG/290-add-js-feature-test-coverage-for-autocomplete .

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -103,6 +103,7 @@ class ResultsView
                    base_query = base_query.where(qualifications: qualifications)
                    base_query = base_query.where(subjects: subject_codes.join(",")) if subject_codes.any?
                    base_query = base_query.where(send_courses: true) if send_courses?
+                   base_query = base_query.where("provider.provider_name" => provider) if provider.present?
 
                    base_query
                      .order("provider.provider_name": results_order)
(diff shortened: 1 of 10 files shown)


dankmitchell:
[[M] provider filter: Do the filtering](https://trello.com/c/UWOeexwH/2957-m-provider-filter-do-the-filtering)

config/settings.yml
```diff
@@ -10,4 +10,4 @@ google:
   places_api_path: "/maps/api/place/autocomplete/json"
 service_support:
   contact_email_address: becomingateacher@digital.education.gov.uk
-redirect_results_to_c_sharp: true
+redirect_results_to_c_sharp: false
```
aliuk2012: Too early to remove this?

gpeng: I think we still need to leave it for now. It does need to be set to `true` though 🤦‍♂

PR #128, opened 2020-02-24
Spike into integrating StandardJS to Lint our JS

aliuk2012:
### Context
JavaScript should be treated like Ruby and SCSS. We should lint it to ensure coding styles and standards are adhered to.

This PR adds Standard JS and fixes issues identified by Standard JS

### DIFF
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -1,38 +1,38 @@
-import "../styles/application.scss";
-import "../images/option-select-toggle-sprite.png";
-import "accessible-autocomplete/dist/accessible-autocomplete.min.css";
-import { initAll } from "govuk-frontend";
-import initLocationsMap from "scripts/locations-map";
-import backLink from "scripts/back-link";
-import initAutocomplete from "scripts/autocomplete";
-import toggle from "scripts/toggle";
-import { loadAnalytics } from "scripts/analytics.js";
-import initCookieBanner from "scripts/cookie-banner";
+import '../styles/application.scss'
+import '../images/option-select-toggle-sprite.png'
+import 'accessible-autocomplete/dist/accessible-autocomplete.min.css'
+import { initAll } from 'govuk-frontend'
+import initLocationsMap from 'scripts/locations-map'
+import BackLink from 'scripts/back-link'
+import initAutocomplete from 'scripts/autocomplete'
+import Toggle from 'scripts/toggle'
+import { loadAnalytics } from 'scripts/analytics.js'
+import InitCookieBanner from 'scripts/cookie-banner'
 
-initAll();
+initAll()
 
-window.initLocationsMap = initLocationsMap;
+window.initLocationsMap = initLocationsMap
 
-const $backLink = document.querySelector('[data-module="back-link"]');
-new backLink($backLink).init();
+const $backLink = document.querySelector('[data-module="back-link"]')
+new BackLink($backLink).init()
 
-const $toggle = document.querySelector('[data-module="toggle"]');
-new toggle($toggle).init();
+const $toggle = document.querySelector('[data-module="toggle"]')
+new Toggle($toggle).init()
 
 initAutocomplete(
   {
-    element: "location-autocomplete",
-    input: "location",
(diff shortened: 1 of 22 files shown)


app/webpacker/scripts/copy-tracking.js
```diff
@@ -1,29 +1,22 @@
...
+CopyTracking.prototype.init = function () {
+  const $module = this.$module
 
   if (!$module) {
-    return;
+    return
   }
 
-  $module.addEventListener("copy", e => {
-    if (event.clipboardData) {
```
aliuk2012: Standard Identified that event was not declared. In this case we should have used `e` which is the passed in value.

defong:
this thing right?

https://standardjs.com/

tvararu:
I initially pushed for standard.js in GDS but now much prefer formatters like [https://prettier.io](https://prettier.io). I have grown to dislike linters that complain about things but don't know how to autofix them.

What are your thoughts about integrating prettier? (Which was used in a non-enforced way to format most of the JS in this project) It also comes with the added bonus of knowing how to format not just JS but also Ruby, Yaml, Markdown, HTML, CSS.

aliuk2012:
This is just a spike to test whether the team would like to use standardJS. I was not planning on not just using Standard JS for code formatting capabilities but more for linting. I can see why you may dislike linters that complain without autofix (Not sure if you mean a feature like `standard --fix` to autocorrect the basic style) but I think for a team that is pretty familiar with linters but not being as familiar with JS it might help detect bugs sooner rather than later and help build confidence. If we went with prettier and decided we wanted to use a linter we would then need to look at eslinter etc. The issue I see with having a setup like that is now we relying on two separate dependencies that require configuration and decisions.


StandardJS, 

> No configuration. The easiest way to enforce code quality in your project. No decisions to make. No .eslintrc files to manage. It just works.
> Automatically format code. Just run standard --fix and say goodbye to messy or inconsistent code.
> Catch style issues & programmer errors early. Save precious code review time by eliminating back-and-forth between reviewer & contributor.

This PR, isn't actually running the standard at this moment but it also picked up a lot of files relying on outside dependencies like GA/Google Maps etc. which can raises bugs if they not initialised and called correctly.

I also like the justification GDS have with regards to why they use StandardJS.https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#why-standard

I'll do another spike for Prettier and then put it to the team to see how they feel about using these tools.

app/webpacker/packs/application.js
```diff
@@ -1,38 +1,38 @@
-import "../styles/application.scss";
```
dankmitchell: Our ruby linter wants double quotes so this will cause confusion...

aliuk2012: I've been thinking about your comment and at first, I thought it might be a deal-breaker.  StandardJS does have an flag `--fix` you could pass to it to fix issues with double quotes not being single. We could wire that up to the rake task as we do for the current ruby linter. 

Do you think its still a deal-breaker?

dankmitchell: I just think we should use double quotes in our JS as well as Ruby. Assuming we can change the config for this linter...

aliuk2012: This is the one of Standard.j

aliuk2012: Unfortunately, we can't change Standard.js rules 

> The whole point of standard is to save you time by avoiding bikeshedding about code style. There are lots of debates online about tabs vs. spaces, etc. that will never be resolved. These debates just distract from getting stuff done. At the end of the day you have to 'just pick something', and that's the whole philosophy of standard -- its a bunch of sensible 'just pick something' opinions. Hopefully, users see the value in that over defending their own opinions.

> Pro tip: Just use standard and move on. There are actual real problems that you could spend your time solving! :P

https://standardjs.com/#why-should-i-use-javascript-standard-style
https://standardjs.com/#i-disagree-with-rule-x-can-you-change-it
https://standardjs.com/#but-this-isnt-a-real-web-standard

defong:
see alive?

aliuk2012:
parking this

PR #148, opened 2020-02-25
[3043] Fix missing filter button copy for mobile users

aliuk2012:
### Context
Mobile users would not be able to see or change filters. This bug is in production.

the Filter button that was displayed for mobile users looked broken without any copy. c# version did have some copy
image

I added the copy which fixed the layout:

image


dankmitchell:
[Fix Filter Button for mobile users](https://trello.com/c/ZmXDb0ad/3043-fix-filter-button-for-mobile-users)

PR #151, opened 2020-02-25
[3044] Add location filtering

asmega:
### Context

- https://trello.com/c/SNDaCBWV/3044-make-find-request-location-filtering
- This hooks up location filtering from backend to the frontend

- Hookup location filtering backend to frontend
- Removed redundant tests, asserting against query parameters
- Rename #distance to #radius

### Guidance to review

- This PR depends on the backend surfacing the `latitude`, `longitude` and `radius` otherwise this feature will not work
- See https://trello.com/c/UkVHbumo/2835-l-location-filter-add-filtering-to-api
- Should wait for that before this card is tested/merged
- Perform a search with any geocoded location e.g. postcode
- Should see results page
- Correct address shows above google map
- Above google map correct radius should show
- Google maps should show with pin in correct location
- Assuming backend works correctly relevant location results should be shown

### DIFF
--- a/spec/features/result_filters/location_spec.rb
+++ b/spec/features/result_filters/location_spec.rb
@@ -69,6 +69,35 @@ feature "Location filter", type: :feature do
     end
   end
 
+  describe "filtering by location" do
+    before do
+      stub_request(:get, courses_url)
+        .with(
+          query: base_parameters.merge("filter[longitude]" => "-0.1300436",
+                                       "filter[latitude]" => "51.4980188",
+                                       "filter[radius]" => "20"),
+        )
+        .to_return(
+          body: File.new("spec/fixtures/api_responses/courses.json"),
+          headers: { "Content-Type": "application/vnd.api+json; charset=utf-8" },
+      )
+    end
+
+    it "displays the courses" do
+      results_page.load
+      results_page.location_filter.link.click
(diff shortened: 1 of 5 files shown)


AgaDufrat:
[[M] Make Find request location filtering](https://trello.com/c/SNDaCBWV/3044-m-make-find-request-location-filtering)

spec/features/result_filters/location_spec.rb
```diff
@@ -69,6 +69,35 @@
     end
   end
 
+  describe "filtering by location" do
```
TobyRet: This was spec was moved out of the "Selecting an option" describe block (line 130). If you think it reads better doing it this way then it would be worth doing the same for the other tests that are still within the "Selecting an option" block for consistency ...

asmega: I've just changed the nesting of those tests. I have a feeling the tests themselves should be changed and not just assert against the query string? Let me know what you think. I'll squash before any merge

TobyRet: Good point. I guess now the API is hooked up we should be asserting what results should appear on the results page also.

asmega: I added a new ticket for the extra work https://trello.com/c/FKnZUeeP/300-refactor-results-filter-tests

PR #159, opened 2020-02-26
[3031] Add a healthcheck endpoint

timabell:
### Context

DevOps

* [3031] Add a healthcheck endpoint

This new `/healthcheck` endpoint checks whether the dependencies of this
service are up. By implication of having returned a 200 OK status you
can also tell that this service is up.

Failure of dependencies results in a 502 bad-gateway status and a json
body showing which dependencies failed.

Front-door will probably continue to check root `/` for a 200 status as
that doesn't have any dependencies currently, so no need for a ping
endpoint as well.

This code is based on
https://github.com/DFE-Digital/teacher-training-api/pull/1202

### Guidance to review

:ship:

### DIFF
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -6,9 +6,7 @@ class HeartbeatController < ActionController::API
         teacher_training_api: api_alive?,
     }
 
-    status = :ok
-    status = :bad_gateway unless checks.values.all?
-    render status: status, json: {
+    render status: (checks.values.all? ? :ok : :bad_gateway), json: {
       checks: checks,
     }
   end
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -1,0 +1,24 @@
+class HeartbeatController < ActionController::API
+  include HTTParty
+
+  def healthcheck
+    checks = {
+        teacher_training_api: api_alive?,
+    }
+
+    status = :ok
+    status = :bad_gateway unless checks.values.all?
+    render status: status, json: {
+      checks: checks,
+    }
+  end
+
+private
+
+  def api_alive?
+    response = HeartbeatController.get("#{Settings.teacher_training_api.base_url}/healthcheck")
+    response.success?
+  rescue StandardError
+    false
+  end
+end
--- a/spec/requests/heartbeat_spec.rb
+++ b/spec/requests/heartbeat_spec.rb
@@ -1,0 +1,81 @@
+require "rails_helper"
+
+describe "GET /healthcheck" do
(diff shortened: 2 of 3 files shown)


timabell:
[Add /healthcheck endpoint to ftt](https://trello.com/c/1jRvOIol/3031-add-healthcheck-endpoint-to-ftt)

app/controllers/heartbeat_controller.rb
```diff
@@ -0,0 +1,24 @@
+class HeartbeatController < ActionController::API
+  include HTTParty
+
+  def healthcheck
+    checks = {
+        teacher_training_api: api_alive?,
+    }
+
+    status = :ok
+    status = :bad_gateway unless checks.values.all?
```
defong:
```suggestion
    status = checks.values.all? ? :ok : :bad_gateway 
```

spec/requests/heartbeat_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(response.body).to eq({ checks: {
+        teacher_training_api: false,
+      } }.to_json)
+    end
+  end
+
+  context "when the api refuses the connection" do
+    before do
+      stub_request(:get, healthcheck_endpoint)
+          .to_raise(StandardError) # https://stackoverflow.com/questions/25552239/webmock-simulate-failing-api-no-internet-timeout/25559291#25559291
```
defong:
```suggestion
          .to_raise(StandardError) 
```

Do think comment adds value

```ruby
stub_request :whatever
  .to_raise :hell
```

spec/requests/heartbeat_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    it "returns the expected response report" do
+      expect(response.body).to eq({ checks: {
+        teacher_training_api: false,
+      } }.to_json)
+    end
+  end
+
+  context "when the api refuses the connection" do
```
defong:
```suggestion
  context "when the api causes exeception" do
```

based on `when the api times out`

aliuk2012: How about `when the api raises an exception`?

timabell: "refuses the connection" tells the reader of the code why we care about the case of an exception being raised

spec/requests/heartbeat_spec.rb
timabell: > Do think comment adds value

yes

app/controllers/heartbeat_controller.rb
timabell: perfect, will test and apply

timabell:
addressed feedback & rebased

PR #165, opened 2020-02-27
[3057] - Fix course count

TobyRet:
NB: this fix requires merging https://github.com/DFE-Digital/teacher-training-api/pull/1207

### Context

- The 'total_courses' helper method provided by the json_api_client gem is an estimate only (it is calculated by multiplying number of pages by the number per page). This results in course count inaccuracies under certain search conditions

- The course count is now determined by the value provided via the meta object in the API response (See https://github.com/DFE-Digital/teacher-training-api/pull/1207). Doing it in this way also follows the JSON API spec recommendations

- Calculate course count using the value provided in the API response meta object
- Update JSON fixture

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -117,7 +117,7 @@ class ResultsView
   end
 
   def course_count
-    courses.total_count
+    courses.meta["count"]
   end
 
   def subjects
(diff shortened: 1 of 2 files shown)


TobyRet:
[Fix results page total count](https://trello.com/c/rjDbbewU/3057-fix-results-page-total-count)

PR #175, opened 2020-03-02
[3067] Replace chrome/selenium with webdrivers gem

aliuk2012:
### Context
When running tests the following warning was being logged:

`
 WARN Selenium [DEPRECATION] Selenium::WebDriver::Chrome#driver_path= is
deprecated. Use Selenium::WebDriver::Chrome::Service#driver_path= instead.
`

The reason is that chrome-drivers has been deprecated since 2019-03-31.
It is recommended to use `webdrivers` gem (Selenium tests more easily
with automatic installation and updates for all supported webdrivers)

Other references:
stackoverflow.com/questions/55970418/capyabra-selenium-chrome-driver-settings

### Guidance to review
Replicates this PR https://github.com/DFE-Digital/govuk-rails-boilerplate/pull/197

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -62,8 +62,6 @@ GEM
       zeitwerk (~> 2.2)
     addressable (2.7.0)
       public_suffix (>= 2.0.2, < 5.0)
-    archive-zip (0.12.0)
-      io-like (~> 0.3.0)
     ast (2.4.0)
     better_errors (2.6.0)
       coderay (>= 1.0.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -310,7 +304,7 @@ GEM
     rubocop-rspec (1.37.1)
       rubocop (>= 0.68.1)
     ruby-progressbar (1.10.1)
-    rubyzip (2.0.0)
+    rubyzip (1.3.0)
```
defong: Is that right?

aliuk2012: Good spot, I did not realise that webdrivers was already on 4.x. I've updated the Gemfile.

aliuk2012:
[Replace chrome/selenium driver gems with webdrivers gem](https://trello.com/c/5iK9RNPH/3067-replace-chrome-selenium-driver-gems-with-webdrivers-gem)

PR #176, opened 2020-03-03
[3064] - Add distance to sort by

TobyRet:
### Context

This PR adds a new 'distance' sort by option on the results page

- Distance sort by option only appears if user is searching by location
- Selecting sort by distance returns results ordered by distance to closest site

- in addition to the above this pr also fixes a bug whereby sort params were being appended to the url rather than mutated

### Guidance to review
- Visit `/location/filter/results` and run a location query (option 1). Once you are on the results page you should see results are automatically sorted by distance. Select other sort options from the drop down and watch the magic show that appears before your eyes.

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -4,6 +4,7 @@ class ResultsView
 
   MAXIMUM_NUMBER_OF_SUBJECTS = 43
   NUMBER_OF_SUBJECTS_DISPLAYED = 4
+  DISTANCE = "2".freeze
 
   def initialize(query_parameters:)
     @query_parameters = query_parameters
@@ -73,6 +74,10 @@ class ResultsView
     query_parameters["rad"]
   end
 
+  def sort_by
+    query_parameters["sortby"]
+  end
+
   def show_map?
     latitude.present? && longitude.present? && radius.present?
   end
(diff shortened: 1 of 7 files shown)


TobyRet:
[[M] Add distance to sort by](https://trello.com/c/lDC56w6Y/3064-m-add-distance-to-sort-by)

app/helpers/view_helper.rb
```diff
@@ -22,4 +22,17 @@ def bat_contact_email_address
   def bat_contact_mail_to(name = nil, subject: nil, link_class: "govuk-link")
     mail_to bat_contact_email_address, name || bat_contact_email_address, subject: subject, class: link_class
   end
+
+  def sort_options
```
gpeng: I think this would make more sense in the `ResultsView` as that already handles the view logic where this is used and it seems a bit strange referencing it from here. It ties this helper to only being used on the `results#index` anyway.

TobyRet: Ok good call

app/view_objects/results_view.rb
```diff
@@ -103,6 +108,23 @@ def provider_filter?
     query_parameters["l"] == "3"
   end
 
+  def sort_by_distance?
+    query_parameters[:sortby] == DISTANCE
```
defong:
```suggestion
    sort_by == DISTANCE
```

is `sort_by` getting used anyways?

TobyRet: Yeah mate good spot. I've changed

PR #179, opened 2020-03-04
[3070] - fix user journey when using the wizard

aliuk2012:
### Context

This fixes a bug where users starting out from the start page
(i.e location page) would be taken straight to the results page if they
search by postcode,town or city. The preferred route was supposed to be
that all location-based search (i.e "search by postcode, town or city"
or "Across the whole of England") should always take the user to the
subjects page to select specific subjects they are interested in.

Provider searches, on the other hand, should take users straight to the
results page.

As the location page (ie "Start page") has a dual purpose if the user
has already searched for some courses but decided to edit the location
then they should not have to go back to the subject page before returning
to the results page.

I found LocationController create method very hard to read because of
all the redirecting and various params being passed around. I understand
that the reason the params are so complicated is that we have to support
a c# version of the service. We should look to review this in more detail
once we are using rails completely.

I made a start at refactoring the method to attempt to make it easier to
read and understand by moving some of the logic out into private methods.

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -13,7 +13,7 @@ module ResultFilters
     def create
       # if searching for specific provider go to results page
       if provider_option_selected?
-        redirect_to(provider_path(params_for_provider_search))
+        redirect_to(provider_path(get_params_for_selected_option({})))
         return
       end
 
@@ -21,14 +21,7 @@ module ResultFilters
       form_object = LocationFilterForm.new(form_params)
       if form_object.valid?
         all_params = form_params.merge!(form_object.params)
-
-        if location_option_selected?
-          submitted_params = params_for_location_search(all_params)
-        elsif across_england_option_selected?
-          submitted_params = params_for_across_england_search(all_params)
-        end
-
-        redirect_to(next_step(submitted_params))
+        redirect_to(next_step(all_params))
       else
         flash[:error] = form_object.errors
         back_to_current_page_if_error(form_params)
@@ -54,23 +47,22 @@ module ResultFilters
       filter_params[:l] == "3"
     end
 
-    def params_for_location_search(params)
-      params.except(:query)
-    end
-
-    def params_for_across_england_search(params)
-      params.except(:lat, :lng, :rad, :loc, :lq, :query, :sortby)
-    end
-
-    def params_for_provider_search
-      filter_params.except(:lat, :lng, :rad, :loc, :lq)
+    def get_params_for_selected_option(all_params)
+      if location_option_selected?
+        all_params.except(:query)
+      elsif across_england_option_selected?
+        all_params.except(:lat, :lng, :rad, :loc, :lq, :query, :sortby)
+      elsif provider_option_selected?
+        filter_params.except(:lat, :lng, :rad, :loc, :lq)
+      end
     end
 
     def strip(params)
       params.reject { |_, v| v == "" }
     end
 
-    def next_step(submitted_params)
+    def next_step(all_params)
+      submitted_params = get_params_for_selected_option(all_params)
       if flash[:start_wizard]
         start_subject_path(submitted_params)
       else
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -11,15 +11,9 @@ module ResultFilters
     end
 
     def create
+      # if searching for specific provider go to results page
       if provider_option_selected?
         redirect_to(provider_path(params_for_provider_search))
-        return
-      elsif across_england_option_selected?
-        if flash[:start_wizard]
-          redirect_to(start_subject_path(params_for_across_england_search))
-        else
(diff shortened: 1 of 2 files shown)


aliuk2012:
[Location search skips subject wizard](https://trello.com/c/UyDr1UL9/3070-location-search-skips-subject-wizard)

app/controllers/result_filters/location_controller.rb
```diff
@@ -11,30 +11,27 @@ def start
     end
 
     def create
+      # if searching for specific provider go to results page
```
TobyRet: This comment needed? Also looks like it's going to provider_path not results_path

app/controllers/result_filters/location_controller.rb
```diff
@@ -45,6 +42,10 @@ def build_results_filter_query_parameters
         .query_parameters_with_defaults
     end
 
+    def location_option_selected?
```
TobyRet: I wonder if it makes sense to add `location_option_selected` and `across_england_selected` methods to the `LocationForm`.

app/controllers/result_filters/location_controller.rb
aliuk2012: I thought it might be handy to know that the provider_path also does a redirect depending on whether the provider name exists and only returns 1 record else it redirects the user to a provider suggestion page for them to select a specific one.

app/controllers/result_filters/location_controller.rb
aliuk2012: I'm not familiar enough to have an opinion on it. What does everyone else think about moving `location_option_selected?` and `across_england_selected?` methods to `LocationForm`

defong: just throwing this into the mist....

3 of those `option_selected?` looks like it should be a switch case statement....

```ruby
    def create
      # if searching for specific provider go to results page
      if filter_params[:l] == "3"
        redirect_to(provider_path(params_for_provider_search))
        return
      end

      form_params = strip(filter_params.clone)
      form_object = LocationFilterForm.new(form_params)
      if form_object.valid?
        all_params = form_params.merge!(form_object.params)

        if filter_params[:l] == "1"
          submitted_params = params_for_location_search(all_params)
        elsif filter_params[:l] == "2"
          submitted_params = params_for_across_england_search(all_params)
        end

        redirect_to(next_step(submitted_params))
      else
        flash[:error] = form_object.errors
        back_to_current_page_if_error(form_params)
      end
    end
```

fofr:
Worth noting: When editing location (from the filters on search) you should come straight back to the results.

app/controllers/result_filters/location_controller.rb
aliuk2012: I think its easier to read if they are private methods but I have taken on board your comments and have pushed up a new commit. It reduces the number of private methods and should make the `create` method easier to read.

aliuk2012: 😭  not sure why the tests are now failing on azure but passing locally

app/controllers/result_filters/location_controller.rb
danielburnley: IMO changing the method `provider_option_selected?` would be much clearer than having a comment - I had a similar response to toby in that I read this comment and saw the redirect and felt like this was an out of place/out of date comment. Would be worth improving the code if you think it's not clear

app/controllers/result_filters/location_controller.rb
```diff
@@ -53,16 +47,35 @@ def provider_option_selected?
       filter_params[:l] == "3"
     end
 
-    def params_for_provider_search
-      filter_params.except(:lat, :lng, :rad, :loc, :lq)
-    end
-
-    def params_for_across_england_search
-      filter_params.except(:lat, :lng, :rad, :loc, :lq, :query)
+    def get_params_for_selected_option(all_params = nil)
```
danielburnley: Is there a reason why we give a default of `nil` here? Feels like stuff is expecting a hash, so i'd expect a default to an empty hash

danielburnley: (if a default is even required)

app/controllers/result_filters/location_controller.rb
```diff
@@ -53,16 +47,35 @@ def provider_option_selected?
...
+      if location_option_selected?
+        all_params.except(:query)
+      elsif across_england_option_selected?
+        all_params.except(:lat, :lng, :rad, :loc, :lq, :query, :sortby)
+      elsif provider_option_selected?
+        filter_params.except(:lat, :lng, :rad, :loc, :lq)
```
danielburnley: Optional: You could refactor these to be a bit clearer as to what you're removing from each hash to the reader, but these param names will hopefully be changed in the future so it's up to your discretion

aliuk2012: I agree with you. I think the params, in general, need to be refactored and reduced but that can only be done once we turn off c# version. I think we should leave this as it is for the moment

PR #180, opened 2020-03-05
[3066] Sitemap is now unpaginated

asmega:
### Context

- https://trello.com/c/9oyEkcDp/3066-s-update-find-sitemapcontroller-to-call-the-api-with-the-correct-parameters-to-disable-pagination
- Previously only a subset of courses would show for sitemap as api call was paginated
- This makes the api call with very large pagination window to return full dataset
- This api endpoint has been optimised to only return the data needed to generate the sitemap 

- Make call to optimised endpoint for unpaginated courses
- Changed `last_published_at` to `changed_at`

### Guidance to review

- `curl http://localhost:3002/sitemap.xml`
- this will make api call and only return needed data
- should return full dataset
- should take under 10 seconds

### DIFF
--- a/app/controllers/sitemaps_controller.rb
+++ b/app/controllers/sitemaps_controller.rb
@@ -1,8 +1,10 @@
 class SitemapsController < ApplicationController
   def show
     @courses = Course
-      .includes(:provider)
       .where(recruitment_cycle_year: Settings.current_cycle)
+      .select("course_code", "provider_code", "changed_at")
+      .page(1)
+      .per(20000)
       .all
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/sitemaps_controller.rb
```diff
@@ -1,8 +1,10 @@
 class SitemapsController < ApplicationController
   def show
     @courses = Course
-      .includes(:provider)
       .where(recruitment_cycle_year: Settings.current_cycle)
+      .select("course_code", "provider_code", "changed_at")
+      .page(1)
+      .per(20000)
```
aliuk2012: What if there's more than 20k. Could this value be dynamic like `Course.count()` or something so its also return all the results on one page

asmega: The backend is hardcoded to 20K. So if the number is over that it'll chop it off (not sure but might go to next page?).

PR #187, opened 2020-03-06
[3084] - Bugfix: Some course results are not displaying nearest site

TobyRet:
### Context
We noticed that some course results had an empty value for 'nearest site' on the results page. 

- When displaying course results we do not want to consider those sites that have empty address details
- If a course has no sites with address details then 'nearest site' and 'distance to nearest' site info should not be displayed
- We now need to request `site_statuses` when calling the API and make sure we only consider sites with an 'new' or 'running' status (this now mirrors the the API call in the Courses controller)
- The results page also now correctly displays site distance depending on the value. If the site distance is 1 mile then the result displays '1 mile' else 'X miles' is displayed

### Guidance to review
- Try and find a course with an empty 'nearest site' field in the course results
- Flick through the results pages and make sure nothing blows up

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Some course results have a missing 'nearest address'](https://trello.com/c/0CDyhtC8/3084-some-course-results-have-a-missing-nearest-address)

app/view_objects/results_view.rb
```diff
@@ -202,6 +203,18 @@ def subjects
 
 private
 
+  def ordered_sites_for(course)
```
misaka: This is called `ordered_sites_for`, but it doesn't sort/order the sites it extracts. How about calling it `new_or_running_sites_for`?

`#namingishard`

app/view_objects/results_view.rb
```diff
@@ -127,9 +127,10 @@ def sort_options
 
   def courses
     @courses ||= begin
-                   base_query = Course
-                     .includes(:provider).includes(:sites)
-                     .where(recruitment_cycle_year: Settings.current_cycle)
+                   base_query = Course.
+                     includes(site_statuses: [:site]).
+                     includes(provider: [:sites]).
```
misaka: Do we need to add the sites to the provider? Or is this some weird JSONAPIClient thing where if we include sites for `site_statuses` we also have to include them for `provider`?

app/view_objects/results_view.rb
TobyRet: 👍

app/view_objects/results_view.rb
TobyRet: 👀

TobyRet: Nah we don't need it. I've removed

PR #196, opened 2020-03-12
[3120] - Bugfix: Strip out non-ASCII characters

TobyRet:
### Context
When a user visits the `/results/filter/location` or `/results/filter/provider` forms we want to make sure we strip out user any input containing non-ASCII characters before the query is sent to the V3 API.

Creation and use of a new controller concern which strips out non-ASCII characters

### Guidance to review
Not sure if it should be part of this PR but this LocationsController is in need of some refactoring methinks.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product review

### DIFF
--- a/app/controllers/concerns/query_sanitizer.rb
+++ b/app/controllers/concerns/query_sanitizer.rb
@@ -1,0 +1,5 @@
+module QuerySanitizer
+  def strip_non_ascii(query)
+    query.encode("ASCII", invalid: :replace, undef: :replace, replace: "")
+  end
+end
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -1,6 +1,7 @@
 module ResultFilters
   class LocationController < ApplicationController
     include FilterParameters
+    include QuerySanitizer
 
     before_action :build_results_filter_query_parameters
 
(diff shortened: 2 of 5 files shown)


app/controllers/concerns/query_sanitizer.rb
```diff
@@ -0,0 +1,5 @@
+module QuerySanitizer
+  def strip_non_ascii(query)
```
TobyRet: Naming stuff is hard. Let me know if you have better alternatives

app/controllers/result_filters/location_controller.rb
```diff
@@ -11,13 +12,17 @@ def start
     end
 
     def create
+      sanitized_params = filter_params.
```
TobyRet: We should strip out non-ASCII characters before redirecting to the ProviderController

app/controllers/result_filters/provider_controller.rb
```diff
@@ -11,22 +12,27 @@ def new
       @provider_suggestions = Provider
         .select(:provider_code, :provider_name)
         .where(recruitment_cycle_year: Settings.current_cycle)
-        .with_params(search: params[:query])
+        .with_params(search: sanitized_params["query"])
```
TobyRet: A user could potentially go straight to the /results/filter/provider controller direct (without being redirected there). Strip out non-ASCII characters again?

app/controllers/concerns/query_sanitizer.rb
defong: is it actually striping it or encoding it :thinking:

app/controllers/result_filters/provider_controller.rb
defong: Maybe on the rack middleware level.... this selective satanizing looks hellish

TobyRet:
[Remove non-ASCII characters from provider-suggestions query](https://trello.com/c/Ou8oRtwC/3120-remove-non-ascii-characters-from-provider-suggestions-query)

app/controllers/concerns/query_sanitizer.rb
TobyRet: Stripping

TobyRet:
Closing this PR as this is an encoding problem. Will raise new PR

PR #200, opened 2020-03-13
[3127] Fix qualification backlink after failed validation

gpeng:
### Context

If the user deselects all qualifications in the qualifications filter, the query parameter is set to `none` and the user is shown a validation failure message. Unfortunately that parameter is also added to the back link on that page. If the user clicks that link, the filter on the results page is empty.

~~To fix this issue `all_qualifications?` returns `true` when the parameter is set to `none`. This is mildly hacky but it would be much more complex to track previous parameters and add logic to manage them so this seems to be a pragmatic approach to this edge case.~~

The parameter has been removed. The results page link passes all three qualifications in the link querystring so the behaviour from there is unaltered. If the user visits the filter directly without querystring params then all three checkboxes will now be unchecked. This is an edge case and the user would be presented with a validation message if they were to try and submit the empty form. I think this behaviour is more 'correct' but it does differ from the old version. 

Remove the `qualifications=none` parameter that was used on failed validation.

### Guidance to review

* Go to the results page
* Click the qualifications filter
* Deselect everything
* Submit
* When you see the validation message, click the back link
* All qualifications should be visible in the filters

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/result_filters/qualification_controller.rb
+++ b/app/controllers/result_filters/qualification_controller.rb
@@ -11,7 +11,7 @@ module ResultFilters
         redirect_to results_path(filter_params)
       else
         flash[:error] = "Please choose at least one qualification"
-        redirect_to qualification_path(filter_params.merge(qualifications: "none"))
+        redirect_to qualification_path(filter_params)
       end
     end
 
--- a/app/view_objects/result_filters/qualification_view.rb
+++ b/app/view_objects/result_filters/qualification_view.rb
@@ -1,7 +1,7 @@
 module ResultFilters
   class QualificationView
     def initialize(params:)
-      @qualifications_parameter = params[:qualifications]
+      @qualifications_parameter = params[:qualifications] || ""
     end
 
     def qts_only_checked?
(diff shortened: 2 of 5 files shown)


gpeng:
[Any Qualification validation error breaks backlink](https://trello.com/c/rYVHWwmV/3127-any-qualification-validation-error-breaks-backlink)

defong:
I do wonder is `qualifications=none` a hack and we just fixed it by building on a hack?

gpeng:
Yeah... it does feel hacky. An alternative would be to not add any QS params. We could maybe set the state on the results filter on that page to `if all_qualifications? || no_qualifications? then...`. At least that would be more explicit.

defong:
Agreed as https://find-teacher-3127-any-q-q70l5q.herokuapp.com/results?qualifications=none
looks like a hopeless qualification....

gpeng:
@defong I've updated this now and got rid of the pesky 'none'. None is gone.

PR #201, opened 2020-03-13
Bump Node.Js 10 to 12.16.1

aliuk2012:
The reason for this bump is that Nodejs 10 is going in maintenance mode
from April 2020. To stay up to date on dependency, now might be a good
time to migrate to version 12.

https://nodejs.org/en/about/releases/

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review


defong:
What's actually holding this up?

Do we need to double check the docker file?

TobyRet:
We will need a similar upgrade for Publish Teacher Training

aliuk2012:
We need to check dockerfile and any where else in our pipeline where we depend on node

aliuk2012:
@TobyRet I'm happy to create a similar pr for publish

PR #222, opened 2020-03-23
[3155] Fix error styling and standardise behaviour across the service

aliuk2012:
## Background
During migration from c# to Rails, a lot of different developers worked on various pages and were mostly trying to replicate what was done in c#. Unfortunately, this introduced some inconsistency in styling, error messages and interaction when the user clicked on the error summary link.

## Changes proposed in this pull request

**Home/Location page**
- Add missing error message when no type of location search is selected [e41e60a]
- reorder the hint/error message so that it matches [govuk-frontend guidelines](https://design-system.service.gov.uk/components/text-input/#error-messages) [40705f1]
- fixes malformed HTML where two elements on the page had the same id. This was done by excluding the keys in the hidden fields. (No visual difference)[d204866]
- fixes an accessibility issue with location/provider labels and text fields were not associated so screenreaders would not be able to provide the context of the input to the user. (No visual difference but users should be able to click on the labels)[e20d44e]
- Error summary link should set focus on the first radio/checkbox in a list or the invalid text field [d986e40]

**Study type filter page**
- Error summary link should set focus on the first radio/checkbox in a list or the invalid text field [d986e40]
- Does not set the default to both full time/parttime if there is an error. This fixes a usability issue where a user may untick full-time and part-time checkboxes, they get an error message saying they need to select at least one but both checkboxes are ticked already.[d8083f0]

**Qualification filter page**
- Error summary link should set focus on the first radio/checkbox in a list or the invalid text field (No visual difference)[04409e2]

#### Before - Home/Location page (3 error states)

image

image
image

#### After - Home/Location page (3 error states)

image

image

image
------
#### Before - Study type filter page 
image

#### After - Study type filter page
image

## Outstanding issues not addressed in this PR
**Home/Location page**
- Accessibility autocomplete does not allow additional css classes to be added to the text field. https://github.com/alphagov/accessible-autocomplete/issues/428 

**Subject page**
- Error summary does not link to the first checkbox because accordion sections do not expand when expected. This is because accordion has a feature that stores the state of each section in session storage. The session storage overrides the state of the accordion that the server has sent.
- Checkboxes do not have error style i.e red left border and indent (I did try to apply it across the whole accordion but it didn't feel like it added much value to the accordion.

Generally, the subject page is complicated because we are using an accordion to group what would have been a large list of checkboxes into different groups.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -13,7 +13,7 @@
         end
     } %>
     <%= form_with url: location_path, method: :post, data: { "ga-event-form" => "Location" } do |form| %>
-      <%= render "shared/hidden_fields", exclude_keys: ["l"], form: form %>
+      <%= render "shared/hidden_fields", exclude_keys: %w(l query rad), form: form %>
       <fieldset class="govuk-fieldset" role="radiogroup" aria-required="true" class="govuk-fieldset">
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
           <h1 class="govuk-heading-xl">Find courses by location or by training provider</h1>
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -18,7 +18,16 @@
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
           <h1 class="govuk-heading-xl">Find courses by location or by training provider</h1>
         </legend>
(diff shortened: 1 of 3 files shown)


aliuk2012:
[[S] Fix error message inconsistency (design review)](https://trello.com/c/nlvmT5ng/3155-s-fix-error-message-inconsistency-design-review)

PR #223, opened 2020-03-23
[3158] Update ruby version to 2.6.5 and nodejs to 12.16.1

AgaDufrat:
### Guidance to review
- `asdf install ruby 2.6.5`
- `asdf install nodejs 12.16.1`
- `bundle install`
- check if the app and tests are running 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review


AgaDufrat:
[Update Ruby everywhere to at least 2.6.5](https://trello.com/c/tvZnwfLX/3158-update-ruby-everywhere-to-at-least-265)

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.6.1-alpine
+FROM ruby:2.6.5-alpine
```
aliuk2012: When I looked at updating node, I noticed that 2.6.5 image uses node 12.

aliuk2012: I think we should try to ensure that we are using the same version of node between what we are deploying and using locally. Perhaps we should bump node version in this pr

AgaDufrat: sure, will give it a go

aliuk2012:
Any thoughts about going straight to Ruby 2.7.0? To soon? I attempted it locally and the tests pass but a few deprecation warnings being raised by some of the gems we are using.

PR #224, opened 2020-03-23
[2927] Add redirect for www2 and education to www service

dankmitchell:
### Context
c# (www) is now dead so we need to promote  our new rails app to www instead of www2

Add redirect for www2 and education.gov.uk to www. service

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,7 +1,7 @@
 # rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
-  constraints(host: /www2\.|\.education\./) do
+  constraints(host: /^www2\.|\.education\./) do
     match "/(*path)" => redirect { |_, req| "#{Settings.base_url}#{req.fullpath}" },
       via: %i[get post put]
   end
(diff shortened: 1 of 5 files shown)


dankmitchell:
[Redirect www2 Find Rails traffic to www.find](https://trello.com/c/OAEZhP9G/2972-redirect-www2-find-rails-traffic-to-wwwfind)

misaka:
I think this should be done via FrontDoor.

config/routes.rb
```diff
@@ -1,6 +1,11 @@
 # rubocop:disable Metrics/BlockLength
 
 Rails.application.routes.draw do
+  constraints(host: /www2\.|\.education\./) do
```
misaka:
```suggestion
  constraints(host: /^www2\.|\.education\./) do
```

In case we were to implement this in the app, should probably lock `www2` for the sake of clarity to the reader if nothing else.

timabell:
:sheep: it

I'd prefer rails to do it just because frontdoor is a black box managed outside the team

PR #225, opened 2020-03-24
[3125] Backlink on location page returns no results after validation error

gpeng:
### Context

When selecting a 'location' filter (or provider filter as they are selected on the same page) there is a back link. We populate the backlink with the values from the original querystring that arrives with the filter page request. When an invalid form is submitted from that page the backlink is rebuilt with the now invalid parameters. If the backlink is clicked a variety of broken results pages are displayed depending on what combination of original and invalid filters are selected. This leads to either an incorrect filter being applied or in some cases no results at all and on '14xxx across England' link being displayed.

Store the parameters from the last valid results page in a set of hidden fields prefixed with `prev_`. In the event of an invalid form being submitted from the location page, use these to build the back link.

Remove these parameters before the results page is rendered both from the back link and also from the provider links on the interstitial (non-javascript) provider list page.

### Guidance to review

**With JS:**

* Create a valid location filter
* Click 'Change location or choose a provider' and submit an empty provider filter form so as an error message is displayed. 
* Click the back link
* You should be returned to the original location filter

* Create a valid provider filter
* Click 'Change location or choose a provider' and submit an empty location filter form so as an error message is displayed. 
* Click the back link
* You should be returned to the original provider filter

**Without JS:**

Do those ^

also...

* Submit an incomplete provider search so as the provider list is displayed. 
* Check that the links for the provider work

In all cases the `prev_` parameters should have been removed by the time you return to the results page.

The start pages don't have a back link but will still get `prev_` parameters in their querystring if an invalid form is submitted. These should have been removed by the time the results page is reached and will have been set to `none` which is ignored.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/concerns/filter_parameters.rb
+++ b/app/controllers/concerns/filter_parameters.rb
@@ -1,11 +1,7 @@
 module FilterParameters
-  def filter_params
-    if request.method == "GET"
-      parameters = request.query_parameters
-    elsif request.method == "POST"
-      parameters = request.request_parameters
-    end
+  PREVIOUS_PARAMETER_PREFIX = "prev_".freeze
 
+  def filter_params
     custom_params = parameters.reject do |param|
       param.in? %w(utf8 authenticity_token)
     end
@@ -25,7 +21,37 @@ module FilterParameters
     params[param].split(",") if params.key?(param)
   end
 
+  def filter_params_without_previous_parameters
+    remove_previous_parameters(filter_params)
+  end
+
+  def merge_previous_parameters(all_parameters)
+    previous_parameters.each do |key, value|
+      next if value == "none"
+
+      all_parameters[key.delete_prefix(PREVIOUS_PARAMETER_PREFIX)] = value
+    end
+
+    remove_previous_parameters(all_parameters)
+  end
+
+private
+
   def serialize_array_filter_param(value)
     value.join(",") if value.present?
+  end
+
+  def parameters
+    return request.query_parameters if request.method == "GET"
+
+    request.request_parameters
+  end
+
+  def previous_parameters
+    parameters.select { |key, _value| key.start_with? PREVIOUS_PARAMETER_PREFIX }
+  end
+
+  def remove_previous_parameters(all_parameters)
+    all_parameters.reject { |key, _value| key.start_with? PREVIOUS_PARAMETER_PREFIX }
   end
 end
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -19,20 +19,22 @@ module ResultFilters
 
       form_params = strip(filter_params.clone).merge(sortby: ResultsView::DISTANCE)
       form_object = LocationFilterForm.new(form_params)
+
       if form_object.valid?
-        all_params = form_params.merge!(form_object.params)
-        redirect_to(next_step(all_params))
+        parameters_with_geocode_added_and_previous_removed = remove_previous_parameters(form_params.merge(form_object.params))
+        redirect_to(next_step(parameters_with_geocode_added_and_previous_removed))
       else
         flash[:error] = form_object.errors
-        back_to_current_page_if_error(form_params)
+        back_to_current_page_if_error(merge_previous_parameters(form_params))
       end
     end
 
   private
 
     def build_results_filter_query_parameters
-      @results_filter_query_parameters = ResultsView.new(query_parameters: request.query_parameters)
-        .query_parameters_with_defaults
+      @results_filter_query_parameters = merge_previous_parameters(
+        ResultsView.new(query_parameters: request.query_parameters).query_parameters_with_defaults,
+      )
     end
 
     def location_option_selected?
--- a/app/controllers/result_filters/provider_controller.rb
+++ b/app/controllers/result_filters/provider_controller.rb
@@ -18,7 +18,9 @@ module ResultFilters
         flash[:error] = [I18n.t("location_filter.fields.provider")]
         redirect_back
       elsif @provider_suggestions.count == 1
-        redirect_to results_path(filter_params.merge(query: @provider_suggestions.first.provider_name))
+        redirect_to results_path(
+          filter_params_without_previous_parameters.merge(query: @provider_suggestions.first.provider_name),
+        )
       end
     end
 
--- a/app/views/result_filters/provider/new.html.erb
+++ b/app/views/result_filters/provider/new.html.erb
@@ -10,7 +10,10 @@
     <ul class="govuk-list govuk-list--bullet">
       <% @provider_suggestions.each do |provider| %>
         <li data-qa="provider_suggestion">
-          <%= govuk_link_to "#{provider.provider_name} (#{provider.provider_code})", results_path(filter_params.merge(query: provider.provider_name)), { class: "govuk-link", data: { qa: "provider_suggestion__link" } } %>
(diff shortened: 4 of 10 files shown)


gpeng:
[[M] backlink on location page returns no results after validation error](https://trello.com/c/Xt2FB5um/3125-backlink-on-location-page-returns-no-results-after-validation-error)

app/controllers/concerns/filter_parameters.rb
```diff
@@ -25,7 +19,33 @@ def deserialize_array_filter_param(param)
     params[param].split(",") if params.key?(param)
   end
 
+  def remove_previous_parameters(all_parameters)
+    all_parameters.except(*previous_parameters.keys)
+  end
+
+  def merge_previous_parameters(all_parameters)
+    previous_parameters.each do |key, value|
+      next if key =~ /prev_/ && value == "none"
```
danielburnley: I find this really hard to parse, you could maybe extract at least the condition to a method that explained what it was doing to make it clearer

app/controllers/concerns/filter_parameters.rb
```diff
@@ -25,7 +19,33 @@ def deserialize_array_filter_param(param)
...
   end
+
+  def parameters
+    return request.query_parameters if request.method == "GET"
+
+    request.request_parameters
+  end
+
+  def previous_parameters
+    parameters.select { |key, _value| key.to_s =~ /^prev_/ }
```
danielburnley: Wonder if there's a nice way to refactor this regex that's in a lot of places too, maybe moving it to a contatnt like `PREVIOUS_PARAMETER_NAME_MATCHER` or something

app/views/result_filters/provider/new.html.erb
```diff
@@ -10,7 +10,10 @@
     <ul class="govuk-list govuk-list--bullet">
       <% @provider_suggestions.each do |provider| %>
         <li data-qa="provider_suggestion">
-          <%= govuk_link_to "#{provider.provider_name} (#{provider.provider_code})", results_path(filter_params.merge(query: provider.provider_name)), { class: "govuk-link", data: { qa: "provider_suggestion__link" } } %>
+          <%= govuk_link_to "#{provider.provider_name} (#{provider.provider_code})", 
+            results_path(remove_previous_parameters(filter_params.merge(query: provider.provider_name))),
```
danielburnley: Might be worth moving this to a helper method - it's quite hard to read as it is here

app/controllers/concerns/filter_parameters.rb
asmega: would it be safer to use `/^prev_/` here instead of `/prev_/`?

app/views/result_filters/provider/new.html.erb
defong: I do agree with @danielburnley 

`remove_previous_parameters` & `filter_params` are both from FilterParameters`

app/views/shared/_hidden_previous_fields.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= form.hidden_field("prev_l", value: params["prev_l"] || params["l"] || "none") %>
```
asmega: why do we have to assign `none`? can't we skip it if it's `nil`? it'll save you from checking for `none` later?

app/views/result_filters/provider/new.html.erb
misaka: How about `filter_params_without_previous` to combine `remove_previous_parameters` and `filter_params`?

app/controllers/result_filters/provider_controller.rb
```diff
@@ -18,7 +18,9 @@ def new
         flash[:error] = [I18n.t("location_filter.fields.provider")]
         redirect_back
       elsif @provider_suggestions.count == 1
-        redirect_to results_path(filter_params.merge(query: @provider_suggestions.first.provider_name))
+        redirect_to results_path(
+          remove_previous_parameters(filter_params.merge(query: @provider_suggestions.first.provider_name)),
```
misaka: Could also be elligible for `filter_params_without_previous` (or similar), as mentioned below in the view.

app/controllers/concerns/filter_parameters.rb
```diff
@@ -25,7 +19,33 @@ def deserialize_array_filter_param(param)
...
 
+  def remove_previous_parameters(all_parameters)
+    all_parameters.except(*previous_parameters.keys)
+  end
+
+  def merge_previous_parameters(all_parameters)
+    previous_parameters.each do |key, value|
+      next if key =~ /prev_/ && value == "none"
+
+      all_parameters[key.split("_").last] = value
```
misaka:
```suggestion
      all_parameters[key.delete_prefix("prev_")] = value
```

To protect against a subtle gotcha in the future, esp with the params being refactored a little bit. Also better communicates intent to reader.

app/controllers/concerns/filter_parameters.rb
misaka:
```suggestion
      next if key.start_with?("prev_") && value == "none"
```

app/controllers/concerns/filter_parameters.rb
misaka: Also maybe worth using the method-for-the-job:

```suggestion
    parameters.select { |key, _value| key.to_s.start_with? "^prev_" }
```

app/controllers/concerns/filter_parameters.rb
```diff
@@ -25,7 +19,33 @@ def deserialize_array_filter_param(param)
     params[param].split(",") if params.key?(param)
   end
 
+  def remove_previous_parameters(all_parameters)
+    all_parameters.except(*previous_parameters.keys)
```
misaka: Minor nit: is it worth doing this?

```suggestion
    all_parameters.reject { |k,v| k.start_with? "prev_" }
```

It's more performant (fwiw here) and less cognitive load (reader doesn't need to think about `previous_parameters`).

app/controllers/result_filters/location_controller.rb
```diff
@@ -19,20 +19,22 @@ def create
 
       form_params = strip(filter_params.clone).merge(sortby: ResultsView::DISTANCE)
       form_object = LocationFilterForm.new(form_params)
+
       if form_object.valid?
-        all_params = form_params.merge!(form_object.params)
-        redirect_to(next_step(all_params))
+        parameters_with_geocode_added_and_previous_removed = remove_previous_parameters(form_params.merge!(form_object.params))
```
misaka: Is the `merge!` here intentional? Does `form_params` get used in the view or something?

app/views/shared/_hidden_previous_fields.html.erb
gpeng: I think I had to do that as if they aren't set on the original request but then the invalid form causes the non-prev parameters to be present in the querystring on the next request, the partial will populate the `prev_` value from the new invalid parameters. `none` is significant as it indicates that there was no parameter in the last valid filter from the results page and stops it being set subsequently.

app/controllers/result_filters/location_controller.rb
gpeng: I don't think it's needed. I've removed it.

gpeng:
@danielburnley @asmega @misaka Thanks for the reviews. I've addressed all of the comments now I think if you would be so kind as to have another look.

PR #233, opened 2020-03-25
[3178] remove blank provider redirect

asmega:
### Context

- Previously the removed redirect logic was added as there was no validation on the filter form
- Therefore users were able to enter invalid states which would have to be handled in the results page
- As there is now validation on the filter form there is no longer a need to handle these invalid states on the results page

- When provider is blank this previously redirected to `Across England`
- This is no longer needed as filter form now has validation to prevent user from entering this odd invalid state
- Should the user enter this invalid state we still search `Across England` by default

### Guidance to review

- perform search with blank provider in the form
- should not be able to reach removed redirect logic as form validation prevents user from entering this invalid state
- should be validation that informs you to enter a provider
- this should work for both javascript and non-javascript versions

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -1,18 +1,5 @@
 class ResultsController < ApplicationController
   def index
     @results_view = ResultsView.new(query_parameters: request.query_parameters)
-
-    if filtering_by_blank_provider?
-      params_hash = request.query_parameters.merge(l: 2)
-      params_hash.delete("query")
-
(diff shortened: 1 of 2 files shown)


asmega:
[QA results redirect loop](https://trello.com/c/JNcnO1Lv/3178-qa-results-redirect-loop)

PR #235, opened 2020-03-25
[3160] Exclude pagination params from filter_parameters

aliuk2012:
### Context
This resolves the following bug:

Theres a large result set shown on the results page and the user has
used pagination to look through the results. They decide to edit some
filters. After changes to the filter the result set is narrowed down but
the results page tries to take them back to the last set pagination.
This is likely to cause confusion for users or raise errors for us to
investigate. For example the result set is only has 3 courses and the
user was on the 6th page of results.
### Guidance to review

Search for courses "Across England". Use the pagination and then edit each filter and test the following each time:

1. Try the back link - It should take you back to the page you were last on
2. Update a filter and click "Continue" to go back to the results page - You should be taken to the first page of results.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/concerns/filter_parameters.rb
+++ b/app/controllers/concerns/filter_parameters.rb
@@ -3,7 +3,7 @@ module FilterParameters
 
   def filter_params
     parameters.reject do |param|
-      param.in? %w(utf8 authenticity_token)
+      param.in? %w(utf8 authenticity_token page)
     end
   end
 
@@ -20,6 +20,8 @@ module FilterParameters
 
     remove_previous_parameters(all_parameters)
   end
+
+private
 
   def parameters
     return request.query_parameters if %w(GET HEAD).include?(request.method)
--- a/spec/controllers/concerns/filter_parameters_spec.rb
+++ b/spec/controllers/concerns/filter_parameters_spec.rb
@@ -28,6 +28,24 @@ RSpec.describe FilterParameters do
   end
 
   describe "#filter_params" do
+    context "GET" do
+      context "when rails parameters are present" do
+        let(:query_parameters) { { "utf8" => true, "authenticity_token" => "token", "test" => "test" } }
+
+        it "they are stripped" do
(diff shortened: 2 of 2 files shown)


aliuk2012:
[current page remembered after changing search filters](https://trello.com/c/kt3AKH1c/3160-current-page-remembered-after-changing-search-filters)

dankmitchell:
Anyway of added a spec for this?

gpeng:
I've added unit tests for this module in https://github.com/DFE-Digital/find-teacher-training/pull/225 Could you rebase from there and a unit test for this behaviour? I'll hopefully merge that shortly as I've addressed the comments. Just waiting on a 👍

dankmitchell:
Looks good. Needs a rebase @aliuk2012

app/controllers/concerns/filter_parameters.rb
```diff
@@ -22,9 +22,12 @@ def merge_previous_parameters(all_parameters)
...
   def parameters
-    return request.query_parameters if %w(GET HEAD).include?(request.method)
-
-    request.request_parameters
+    request_params = if %w(GET HEAD).include?(request.method)
+                       request.query_parameters
+                     else
+                       request.request_parameters
+                     end
+    request_params.except("page") # exclude pagination params
```
gpeng: Could we do this in the `filter_params` method where we're already filtering out some unwanted parameters?

app/controllers/concerns/filter_parameters.rb
```diff
@@ -22,9 +22,12 @@ def merge_previous_parameters(all_parameters)
   end
```
gpeng: The methods below here are all private. I think the private was removed in https://github.com/DFE-Digital/find-teacher-training/commit/28bc26a1b49c1bffaba9d9ff18fda91c59c8bdc4#diff-6b93dd519c02510c57309a985518c560 possibly by mistake. Could we put it back in?

PR #245, opened 2020-03-31
[3212] Handle subjects as comma delimited string

asmega:
### Context

- https://trello.com/c/QJ6cXQZU/3212-sentry-error-when-subjects-is-a-string-eg-12
- There fixes an error where users are accessing results with subjects as comma delimited string
- They may be an old style from csharp world

- Handle conversion of comma delimited string of subjects

### Guidance to review

- visit url seen in sentry. see trello ticket
- previously would throw an error. should now no longer seen an error and see filter page
- can now submit filter page to see results

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product review


asmega:
[sentry error when subjects is a string e.g "1,2"](https://trello.com/c/QJ6cXQZU/3212-sentry-error-when-subjects-is-a-string-eg-12)

PR #226, opened 2020-03-24
[3028] Add search suggestions when searching with salary

danielburnley:
### Context

Search suggestions need to appear when searching with salary and getting < 3 results

### Changes proposed in this pull request

- Extend current search suggesion logic to encompass salaried searches

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally
- [ ] Product review

PR #251, opened 2020-04-06
[3029] copy update search suggestion when searching for a salaried course when there are 3 salaried courses

gpeng:
### Context

If a user search for salaried courses returns 3 or less results we show them suggested search links. The first link should include salaried and non-salaried courses at a radius that returns more courses than the original search. If widening the radius further returns more results we should show a link to that too. The subsequent links should respect the original filter and only include salaried courses.

If an increment of the radius only returns the same number of courses as the previous radius we don't show the link.

Update the text of the links to better describe the searches that they link to.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/view_objects/suggested_search_link.rb
+++ b/app/view_objects/suggested_search_link.rb
@@ -1,16 +1,17 @@
 class SuggestedSearchLink
   include ActionView::Helpers::TextHelper
-  attr_reader :radius, :count, :parameters, :including_non_salaried
+  attr_reader :radius, :count, :parameters, :including_non_salaried, :explicit_salary_filter
 
-  def initialize(radius:, count:, parameters:, including_non_salaried: false)
+  def initialize(radius:, count:, parameters:, including_non_salaried: false, explicit_salary_filter: false)
     @radius = radius
     @count = count
     @parameters = parameters
     @including_non_salaried = including_non_salaried
+    @explicit_salary_filter = explicit_salary_filter
   end
 
   def text
-    count_prefix = "#{pluralize(count, 'course')} #{'with or without a salary ' if including_non_salaried}"
+    count_prefix = "#{pluralize(count, text_course_text)} "
     count_prefix + (all_england? ? "across England" : "within #{radius} miles")
   end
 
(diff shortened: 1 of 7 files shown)


app/view_objects/suggested_search_link.rb
```diff
@@ -34,4 +35,11 @@ def suggested_search_link_parameters(radius:)
       .reject { |k, _v| %w(lat lng rad loc lq).include?(k) }
       .merge("l" => 2)
   end
+
+  def text_course_text
+    return "non-salaried course" if including_non_salaried
```
defong, line 40: Not too sure about the wording...

non-salaried

TobyRet:
[Copy update: Search suggestion when searching for a salaried course when there are <3 salaried courses](https://trello.com/c/RJNa3CMp/3029-copy-update-search-suggestion-when-searching-for-a-salaried-course-when-there-are-3-salaried-courses)

app/view_objects/suggested_search_link.rb
TobyRet: If Paul Genders is happy, then I'm happy

defong: pretty sure that terminology was leaked 

https://github.com/DFE-Digital/find-teacher-training/pull/226#discussion_r398030780

defong: image

Until this is bottom out it, it is not going anywhere far...


`non-salaried courses` does not mean `Courses with and without salary`
The suggested search number does not tally up/follow through 

LHS url
https://s121d02-3029-find-as.azurewebsites.net/results?fulltime=false&funding=8&hasvacancies=true&l=1&lat=55.953252&lng=-3.188267&loc=Edinburgh%2C+UK&lq=Edinburgh%2C+UK&parttime=false&qualifications[]=QtsOnly&qualifications[]=PgdePgceWithQts&qualifications[]=Other&rad=5&senCourses=false&sortby=2

gpeng: I think this is probably a bug in one of the queries rather than an issue with copy or semantics there of. I'll do another card to look at this so as we can just get the copy sorted in this PR.

gpeng:
image


@defong We've updated some functionality and split a bit of the text out of the link. Could you have another look before we merge? I'll create a new card for the count mismatch bug.

PR #286, opened 2020-05-06
[3332] - Add bad query encoding middleware

TobyRet:
### Context
- Fix for Sentry issue: https://sentry.io/organizations/dfe-bat/issues/?project=1780060&referrer=slack
- Some requests are being made to the 'location-suggestions' endpoint containing invalid encodings. Rails does not handle this natively and exceptions bubble up

- This new middleware intercepts the request **if** the request path == `/location-suggestions` and drops the query if it contains invalid encodings
- Inspired by: https://jamescrisp.org/2018/05/28/fixing-invalid-query-parameters-invalid-encoding-in-a-rails-app/

### Guidance to review
- To view the error locally, comment out the `bad_query_encoding` initialiser, and make a request to `/location-suggestions` with a query containing invalid encodings e.g. `BLAH%%fjf%%BLAH` 
- To view the fix, uncomment the initialiser, restart the app, and send the same request.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/location_suggestions_controller.rb
+++ b/app/controllers/location_suggestions_controller.rb
@@ -1,6 +1,14 @@
 class LocationSuggestionsController < ApplicationController
   def index
+    return render(json: { error: "Bad request" }, status: :bad_request) if params_invalid?
+
     suggestions = LocationSuggestion.suggest(params[:query])
     render json: suggestions
+  end
+
+private
+
+  def params_invalid?
+    params[:query].nil?
   end
 end
--- a/config/initializers/handle_bad_encoding.rb
+++ b/config/initializers/handle_bad_encoding.rb
@@ -1,0 +1,3 @@
+require "rack/handle_bad_encoding"
+
+Rails.application.config.middleware.use Rack::HandleBadEncoding
--- a/lib/rack/handle_bad_encoding.rb
+++ b/lib/rack/handle_bad_encoding.rb
@@ -1,0 +1,19 @@
+module Rack
+  class HandleBadEncoding
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      if env["REQUEST_PATH"] == "/location-suggestions"
+        begin
+          Rack::Utils.parse_nested_query(env["QUERY_STRING"].to_s)
+        rescue Rack::Utils::InvalidParameterError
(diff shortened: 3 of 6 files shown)


TobyRet:
[Entering % in the provider/location suggest fields throws an error](https://trello.com/c/452Zxn8y/3332-entering-in-the-provider-location-suggest-fields-throws-an-error)

defong:
I am unsure of this change, it feels rather heavy handed and may cause cascading issues.

- It is not just this endpoint that it solves/affects `location-suggestions` but across the entirety of the rack.

- Are we now masking the issue, or evolving it to be a bigger issue

ie.
is it now send empty at https://github.com/DFE-Digital/find-teacher-training/blob/116cd5c7a558c50ab7dedee9c466642ca26d84d8/app/services/location_suggestion.rb#L10

deferring to @misaka

TobyRet:
> I am unsure of this change, it feels rather heavy handed and may cause cascading issues.
> 
> * It is not just this endpoint that it solves/affects `location-suggestions` but across the entirety of the rack.
> * Are we now masking the issue, or evolving it to be a bigger issue
> 
> ie.
> is it now send empty at
> 
> https://github.com/DFE-Digital/find-teacher-training/blob/116cd5c7a558c50ab7dedee9c466642ca26d84d8/app/services/location_suggestion.rb#L10
> 
> deferring to @misaka

We could scope this by request path if that was preferred @defong @misaka 

i.e. `"REQUEST_PATH"=>"/location-suggestions"`

Slack discussion - https://ukgovernmentdfe.slack.com/archives/CAHAJFR0X/p1588759652196000

TobyRet:
@defong @misaka 
Do you think it's fine with unit test only or do you think a request spec is necessary as well?

defong:
>    Are we now masking the issue, or evolving it to be a bigger issue

Is it going to send empty?

TobyRet:
> > Are we now masking the issue, or evolving it to be a bigger issue
> 
> Is it going to send empty?

Yes. If a query has invalid encoding (e.g. %%) when it reaches `LocationSuggestion` query will be `nil`. 

empty_request

PR #318, opened 2020-05-28
[3434] - Location filter provider search with one character

TobyRet:
### Context
API has been updated to allow for two character provider searches. See https://github.com/DFE-Digital/teacher-training-api/pull/1378

- On the location filter page, a validation error is displayed on the form if the user searches for a provider with a query string containing only one character

errors

### Guidance to review
- Run this branch against https://github.com/DFE-Digital/teacher-training-api/pull/1378
- Visit `/results/filter/location`. Select other provider and enter a single character in the search field. Validation error should be displayed

### Notes
- Need some copy guidance on the error message
- In doing this work I noticed that GOVUK form builder needs to be added to this page. Also the ProviderController should be using a form object. Potentially the ProviderController logic should be merged into the LocationController and a similar pattern to Publish be adopted. I think this should be a separate piece of work - will add ticket to ideas board.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/result_filters/provider_controller.rb
+++ b/app/controllers/result_filters/provider_controller.rb
@@ -4,7 +4,12 @@ module ResultFilters
 
     def new
       if params[:query].blank?
-        flash[:error] = [I18n.t("location_filter.fields.provider")]
+        flash[:error] = [t("location_filter.fields.provider"), t("location_filter.errors.blank_provider")]
+        return redirect_back
+      end
+
+      if params[:query].length == 1
+        flash[:error] = [t("location_filter.fields.provider"), t("location_filter.errors.invalid_provider")]
         return redirect_back
       end
 
@@ -15,7 +20,7 @@ module ResultFilters
         .all
 
       if @provider_suggestions.count.zero?
-        flash[:error] = [I18n.t("location_filter.fields.provider")]
+        flash[:error] = [I18n.t("location_filter.fields.provider"), I18n.t("location_filter.errors.missing_provider")]
         redirect_back
       elsif @provider_suggestions.count == 1
         redirect_to results_path(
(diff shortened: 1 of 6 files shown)


TobyRet:
[[S] - Error handling for 1 character provider searches](https://trello.com/c/55f2oRuQ/3434-s-error-handling-for-1-character-provider-searches)

app/controllers/result_filters/provider_controller.rb
```diff
@@ -4,7 +4,12 @@ class ProviderController < ApplicationController
 
     def new
       if params[:query].blank?
-        flash[:error] = [I18n.t("location_filter.fields.provider")]
+        flash[:error] = [I18n.t("location_filter.fields.provider"), I18n.t("location_filter.errors.blank_provider")]
```
AgaDufrat, line 7: I think it would work without `I18n`,  which would make the lines shorter.

```suggestion
        flash[:error] = [t("location_filter.fields.provider"), t("location_filter.errors.blank_provider")]
```
It's a massive nitpick 🙈 and I don't have a strong preference for either so definitely not a blocker.

asmega: it should be noted at this point the 2 are not identical so aren't always interchangeable. https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/translation_helper.rb#L19

asmega: But in this case is probably fine to change

app/helpers/result_filters/location_helper.rb
```diff
@@ -1,9 +1,13 @@
 module ResultFilters
   module LocationHelper
     def provider_error?
+      # binding.pry
```
asmega, line 4: delete me

asmega:
Swapping this out a for a form object i think would be good in the future to remove all the controller validations

PR #327, opened 2020-06-05
[3409] Set cache control in assets header

aliuk2012:
### Context
None of our assets were being cached by the browser because cache-control
was not being set. I set it to a 1 year because our assets use a cache
busting technique so if we were to update their content the filename would
be different and therefore the browser would fetch and use the updated
assets.

Reference:
mikerogers.io/2019/11/02/how-to-set-cache-control-headers-for-rails-webpacker-and-sprockets-assets.html

### Guidance to review
visit https://s121d02-3409-find-as.azurewebsites.net/
Check asset headers and see if cache control is set to expire in a years time 

image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review

### DIFF
--- a/config/initializers/assets_cache_headers.rb
+++ b/config/initializers/assets_cache_headers.rb
@@ -1,0 +1,8 @@
+# This will affect assets served from /app/assets
+Rails.application.config.static_cache_control = "public, max-age=31536000"
+
+# This will affect assets in /public, e.g. webpacker assets.
+Rails.application.config.public_file_server.headers = {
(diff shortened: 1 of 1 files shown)


aliuk2012:
[[M] Cache our assets like stylesheets, fonts etc.](https://trello.com/c/Ef7FCzpy/3409-m-cache-our-assets-like-stylesheets-fonts-etc)

PR #343, opened 2020-06-11
[3524] Smart quotes handles nil input

asmega:
### Context

- https://trello.com/c/FKlI0Stn/3524-sentry-error-nil-for-smart-quotes
- An sentry error was reporting an error around smart quotes when the input was `nil`

- For smart quotes method had nil input and return empty string

### Guidance to review

- test should pass
- alternatively visit sentry error and run same query on local environment and should not see an error

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,6 +20,8 @@ module ApplicationHelper
   end
 
   def smart_quotes(string)
+    return "" if string.blank?
+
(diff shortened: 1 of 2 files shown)


asmega:
[sentry error: nil for smart quotes](https://trello.com/c/FKlI0Stn/3524-sentry-error-nil-for-smart-quotes)

PR #360, opened 2020-07-10
3539 l remove radius from search by location

avinhurry:
### Trello Ticket

https://trello.com/c/5Sud6IFs/3539-l-remove-radius-from-search-by-location

### Context

Users don't need to select a radius, and when they select a small radius of 5 and 10 miles they often filter themselves into a hole – which can lead to the wrong impression that there are no courses near them (eg 0 courses within 5 miles, but actually a good one 5.1 miles away).

- 79% of searches are 20-50 miles
- When users edit the radius, they most commonly switch to a wider radius of 50 miles

## Success

- Remove the radius option from location search
- Default to 50 miles
- Remove the sort-by control for location search, replacing with an indicator that results are sorted by distance
- Change "23 courses found" to "23 courses found within 50 miles of you"

## Tech Notes

* We will want support bookmarked results that use the location filter. We can still only support searching within 50 miles, just ignoring any location params, we just shouldn't blow up if they're included.
* We don't need to worry about bookmarked pages that have a radius param of 100 and are on a page that would no longer give us results for a 50 mile radius search.
* We don't need to support radius in API either so will be able to remove that.

### Things to do
* remove all the radius based suggested searches
* redirect all non 50 radius search to 50 radius search on page 1
* retain only have across England

##### Sorted by distance

image

##### Removed radius

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -64,8 +64,6 @@
                 %>
                 <div id="location-autocomplete" class="govuk-body"></div>
               </div>
-          <%= form.hidden_field(:rad, value: 50) %>
-
             </div>
             <div class="govuk-radios__item">
               <%= form.radio_button(
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -64,24 +64,8 @@
                 %>
                 <div id="location-autocomplete" class="govuk-body"></div>
               </div>
-              <div class="govuk-form-group">
-                <%= form.label :rad, "Within", class: "govuk-label" %>
-                <%= form.select :rad,
-                                options_for_select(
-                                  [
-                                    ["5 miles", "5"],
-                                    ["10 miles", "10"],
-                                    ["20 miles", "20"],
-                                    ["50 miles", "50"],
-                                    ["100 miles", "100"],
-                                  ],
-                                  [request.params[:rad] || "20"]
-                                ),
-                                {include_blank: false},
-                                data: {qa: 'search-radius'},
-                                class: 'govuk-select govuk-!-width-one-half'
-                %>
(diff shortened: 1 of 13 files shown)


defong:
[[L] Remove radius from search by location](https://trello.com/c/5Sud6IFs/3539-l-remove-radius-from-search-by-location)

app/views/result_filters/location/_form.html.erb
```diff
@@ -64,24 +64,8 @@
...
-                                    ["100 miles", "100"],
-                                  ],
-                                  [request.params[:rad] || "20"]
-                                ),
-                                {include_blank: false},
-                                data: {qa: 'search-radius'},
-                                class: 'govuk-select govuk-!-width-one-half'
-                %>
-              </div>
+          <%= form.hidden_field(:rad, value: 50) %>
```
aliuk2012, line 67: Do we need this? Could we not just pass this in the call form find to the teacher-training-api?

spec/site_prism/page_objects/page/results.rb
```diff
@@ -104,6 +104,8 @@ class SuggestedSearchLinkSection < SitePrism::Section
       sections :suggested_search_links, SuggestedSearchLinkSection, '[data-qa="suggested_search_link"]'
 
       section :sort_form, SortFormSection, '[data-qa="sort-form"]'
+
+      element :sorted_by_distance, 'div.govuk-grid-column-full>p.govuk-body.govuk-\\!-margin-bottom-0', text: "Sorted by distance"
```
aliuk2012, line 108: Could we not simplify this selector?

defong: sorted

PR #372, opened 2020-07-23
[3684] Assigns Request_ID to tag in Sentry

JonF1:
### Context
Request ids generated from Find were not showing up in the tags in the error logs in Sentry
[https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find](url)

Request ids are now assigned to tags section in Sentry

### Checklist

- [X] Make sure all information from the Trello card is in here
- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history
- [X] Tested by running locally
- [ ] Product review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,13 +2,14 @@
 
 class ApplicationController < ActionController::Base
   before_action :store_request_id
+  before_action :assign_sentry_contexts
 
   def store_request_id
     RequestStore.store[:request_id] = request.uuid
   end
 
   def assign_sentry_contexts
-    Raven.extra_context(request_id: RequestStore.store[:request_id])
+    Raven.tags_context(request_id: RequestStore.store[:request_id])
   end
 
   def append_info_to_payload(payload)
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -17,6 +17,15 @@ describe ApplicationController, type: :controller do
     end
   end
 
+  describe "#assign_sentry_contexts" do
+    it "assigns a request id to Sentry tags" do
+      allow(Raven).to receive(:tags_context).and_return({})
+      controller.__send__(:store_request_id)
+      controller.__send__(:assign_sentry_contexts)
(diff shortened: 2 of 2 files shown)


JonF1:
[Add `request_id` to logging for Find](https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find)

app/controllers/application_controller.rb
```diff
@@ -2,13 +2,14 @@
 
 class ApplicationController < ActionController::Base
   before_action :store_request_id
+  before_action :assign_sentry_contexts
```
misaka, line 5: Is this testable?

misaka: Just had a look and I don't think this is tested in Publish, but I wonder if it wouldn't be worth testing ... a request spec should trigger this callback and, although it's messy, you should be able to test `Raven.tags_context`. Worth doing?

app/controllers/application_controller.rb
```diff
@@ -2,13 +2,14 @@
...
 
   def store_request_id
     RequestStore.store[:request_id] = request.uuid
   end
 
   def assign_sentry_contexts
-    Raven.extra_context(request_id: RequestStore.store[:request_id])
+    Raven.tags_context(request_id: RequestStore.store[:request_id])
```
misaka, line 12: I don't see a test for this. Is it testable?

spec/controllers/application_controller_spec.rb
```diff
@@ -17,6 +17,14 @@
     end
   end
 
+  describe "#assign_sentry_contexts" do
+    it "assigns a request id to Sentry tags" do
+      expect(Raven).to receive(:tags_context).with(request_id: request_uuid)
+      controller.__send__(:store_request_id)
+      controller.__send__(:assign_sentry_contexts)
```
defong, line 24: can we move the it to sit above the `expect` to make it easier to read

app/controllers/application_controller.rb
defong: it has coverage

app/controllers/application_controller.rb
defong: it has tests

PR #374, opened 2020-07-27
[3684] Add X-Request-Id to API client connection

asmega:
### Context

- https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find
- Find API client was never adding custom `X-Request-Id` header to api calls

- Previous connection_class was not correctly set therefore the custom request_id header was never sent

### Guidance to review

- Make API call in review app e.g. https://s121d02-3684-find-as.azurewebsites.net/results?l=2&subjects%5B%5D=31
- Find the request in kibana
- This request will have an associated `request_id`
- Search for this `request_id` in kibana under the api
- It should have correlating log which will be the corresponding API request

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/spec/models/base_spec.rb
+++ b/spec/models/base_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Base do
+  describe "connection_class" do
+    before do
+      RequestStore.store[:request_id] = "some-request-id"
+    end
+
+    it "connection sends X-Request-Id header" do
+      stub = stub_request(:post, "http://localhost:3001/api/v3/bases")
+        .with(
+          body: "{\"data\":{\"type\":\"bases\",\"attributes\":{}}}",
+          headers: {
+            "X-Request-Id" => "some-request-id",
+          },
+        ).to_return(status: 200, body: "", headers: {})
+
+      Base.new.save
+
+      expect(stub).to have_been_requested
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


asmega:
[Add `request_id` to logging for Find](https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find)

spec/models/base_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it "connection sends X-Request-Id header" do
+      stub = stub_request(:post, "http://localhost:3001/api/v3/bases")
+        .with(
+          body: "{\"data\":{\"type\":\"bases\",\"attributes\":{}}}",
+          headers: {
+            "Accept" => "application/vnd.api+json",
+            "Accept-Encoding" => "gzip,deflate",
+            "Content-Type" => "application/vnd.api+json",
+            "User-Agent" => "Faraday v0.17.3",
```
defong, line 17: Does this mean every update we need to bump this as well?

asmega: At the moment yes, until someone reads the webmock docs and figures out how to only check the one header we need

asmega: I've fixed it now, turned out it was way easier than i thought

PR #401, opened 2020-08-20
[1971] - Return html format from CoursesController#Show

TobyRet:
### Context
- Sentry error: https://sentry.io/organizations/dfe-bat/issues/1816969756/?project=1780060&referrer=slack
- A request to "/course/fonts/Roboto-Regular.ttf" was causing an error to occur because Rails is trying to return the 'not found' page with the wrong content type. It is trying to infer the content type from the bad request, which in this case was `ttf`

- The controller is now hard coded to return in html format.

### Guidance to review

Not entirely sure if this is the best solution to this problem

#### Before

before

#### After

after

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -12,7 +12,7 @@ class CoursesController < ApplicationController
       .find(params[:course_code])
       .first
   rescue JsonApiClient::Errors::NotFound
-    render file: "errors/not_found", status: :not_found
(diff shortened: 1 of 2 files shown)


TobyRet:
https://trello.com/c/xtxdpCey/1971-investigate-sentry-errors

tvararu (approved):

PR #411, opened 2020-08-25
Create better errors experience - redirect to 422 page when query is invalid

TobyRet:
### Context
- A large part of find involves building a search query which is then passed on to the Teacher Training API.
- Previously we were sanitising an invalid query, returning all results. Often these errors are triggered by bots and other _evil_ forces 😈 . But rather than being helpful, this may in fact confuse the user who (unmaliciously) passes in an invalid query by accident (e.g. copy and pasting from a word document).
- Example Sentry error - https://sentry.io/organizations/dfe-bat/issues/1778885702/?environment=production&project=1780060&query=is%3Aunresolved

- The middleware has been changed to redirect any invalid requests to the 422 page.
- Update guidance on both the 422 and 404 pages (now more closely in line with Apply's error pages).

### Guidance to review
- Spin up the app locally and modify your query so that it is invalid. You should be redirected to the 422 page
E.g. `http://localhost:3002/results?%%%%%%%%nasty%%%%%%%%l=2&subjects%5B%5D=31`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/lib/rack/handle_bad_encoding.rb
+++ b/lib/rack/handle_bad_encoding.rb
@@ -5,11 +5,11 @@ module Rack
     end
 
     def call(env)
-      if %w[/location-suggestions /provider-suggestions].include?(env["REQUEST_PATH"])
+      if %w[/location-suggestions /provider-suggestions /results].include?(env["REQUEST_PATH"])
         begin
           Rack::Utils.parse_nested_query(env["QUERY_STRING"].to_s)
         rescue Rack::Utils::InvalidParameterError
-          env["QUERY_STRING"] = ""
+          return [301, { "Location" => "/422", "Content-Type" => "text/html" }, []]
         end
       end
 
--- a/app/views/errors/not_found.html.erb
+++ b/app/views/errors/not_found.html.erb
@@ -4,10 +4,16 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-xl">Page not found</h1>
-      <p class="govuk-body">If you entered a web address please check it was correct.</p>
       <p class="govuk-body">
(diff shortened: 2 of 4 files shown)


TobyRet:
https://trello.com/c/IXmeexZt/1978-find-site-errors

TobyRet:
[Find: Site errors](https://trello.com/c/IXmeexZt/1978-find-site-errors)

lib/rack/handle_bad_encoding.rb
```diff
@@ -5,11 +5,11 @@ def initialize(app)
     end
 
     def call(env)
-      if %w[/location-suggestions /provider-suggestions].include?(env["REQUEST_PATH"])
+      if %w[/location-suggestions /provider-suggestions /results].include?(env["REQUEST_PATH"])
         begin
           Rack::Utils.parse_nested_query(env["QUERY_STRING"].to_s)
         rescue Rack::Utils::InvalidParameterError
-          env["QUERY_STRING"] = ""
+          return [301, { "Location" => "/404", "Content-Type" => "text/html" }, []]
```
tvararu, line 12: This looks like 301 redirect to the /404 page, does that mean that a user would "lose" the page which caused the error once they hit it? Would that be annoying to deal with?

TobyRet: Based on the errors we have seen on Sentry most of these types of invalid queries result from:

1. Users copy and pasting urls containing invalid param (e.g non-ASCII characters). 
2. Bots or malicious users playing around with params

So, in these cases there wasn't a previous page to lose ...

But yes if a user somehow managed to create an invalid query via a form (i.e. entering an invalid postcode, address or provider here https://www.qa.find-postgraduate-teacher-training.service.gov.uk/) then they would lose the page but 
have not seen evidence of this happening (yet).

tvararu (approved):

PR #412, opened 2020-08-26
'Cycle has ended' interim page

TobyRet:
### Context
Find is to be closed on 3rd October (when UCAS stop allowing applications) and will display the interim 'cycle has ended' page.

- Add a new 'cycle has ended' interim page for end of cycle.
- When `Settings.cycle_has_ended` is set to `true`, any requests to the root, search, results and course pages will be redirected to the interim page

### Guidance to review
- In your `development.local.yml` file, set `cycle_has_ended` to `true`.
- When you try to navigate to root, search, results or course pages, you should be redirected to the interim page

### Trello card
https://trello.com/c/8kQxqDWt/2016-dev-find-cycle-has-ended-interim-page-3oct

### Screenshot
cycle_has_ended

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -8,6 +8,16 @@ Rails.application.routes.draw do
   get :healthcheck, controller: :heartbeat
   get :sha, controller: :heartbeat
 
+  if Settings.cycle_has_ended
+    get "/cycle-has-ended", to: "pages#cycle_has_ended", as: "cycle_has_ended"
+    get "/", to: redirect("/cycle-has-ended", status: 302)
+    get "/results", to: redirect("/cycle-has-ended", status: 302), as: "results_redirect"
+    get "/course/*path", to: redirect("/cycle-has-ended", status: 302)
+    get "/start/*path", to: redirect("/cycle-has-ended", status: 302)
+  else
+    get "/cycle-has-ended", to: redirect("/", status: 301)
+  end
+
   scope module: "result_filters" do
     root to: "location#start"
   end
@@ -16,7 +26,7 @@ Rails.application.routes.draw do
     get "subject", to: "subject#start", as: "start_subject"
   end
 
-  # There is no need for a seperate path for start#location but this was the
+  # There is no need for a separate path for start#location but this was the
   # root path in the legacy c# app so we're redirecting to root
   get "/start/location", to: redirect("/", status: "301")
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -6,4 +6,6 @@ class PagesController < ApplicationController
   def privacy; end
 
   def accessibilty; end
+
+  def cycle_has_ended; end
 end
--- a/app/views/pages/cycle_has_ended.html.erb
+++ b/app/views/pages/cycle_has_ended.html.erb
@@ -1,0 +1,17 @@
+<%= content_for :page_title, "Find postgraduate teacher training" %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">Find postgraduate teacher training</h1>
+      <h2 class="govuk-heading-m">Applications closed</h2>
+      <p class="govuk-body">
+        Applications for this academic year (2020 to 2021) are now closed.
+      </p>
+      <h2 class="govuk-heading-m">Find courses starting next academic year (2021 to 2022)</h2>
+      <p class="govuk-body">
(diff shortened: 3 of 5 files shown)


tvararu (approved):

config/routes.rb
```diff
@@ -23,6 +23,7 @@
   get "/terms-conditions", to: "pages#terms", as: "terms"
   get "/accessibility", to: "pages#accessibility", as: "accessibility"
   get "/privacy-policy", to: "pages#privacy", as: "privacy"
+  get "/cycle_has_ended", to: "pages#cycle_has_ended", as: "cycle_has_ended"
```
tijmenb, line 26:
```suggestion
  get "/cycle-has-ended", to: "pages#cycle_has_ended", as: "cycle_has_ended"
```

lib/rack/display_cycle_has_ended_page.rb
```diff
@@ -0,0 +1,29 @@
...
+  class DisplayCycleHasEndedPage
+    include Rails.application.routes.url_helpers
+
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      if cycle_has_ended? && !request_to_cycle_has_ended_page?(env)
+        [301, { "Location" => cycle_has_ended_path, "Content-Type" => "text/html" }, []]
```
tijmenb, line 11: This should probably be a 302, since the redirect is only temporary. This will prevent search engines from updating their index.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

config/routes.rb
tijmenb: To avoid dropping down to using middleware, we could consider something like this:

```rb
if Settings.cycle_has_ended
  get "/cycle-has-ended", to: "pages#cycle_has_ended", as: "cycle_has_ended"
  get "/*", redirect("/cycle-has-ended")
else
  get "/cycle-has-ended", redirect("/")
end
```

tijmenb (commented):
Do we need to redirect all of the pages? Perhaps we need the ping/healtcheck/sitemap pages, as well as the routes to the error controller. Perhaps we can scope this to only the search functionality.

fofr:
Any reason we moved away from the content we used last year?

https://bat-design-history.netlify.app/find-teacher-training/end-of-cycle/

Mainly I prefer "Courses starting in the academic year",

So maybe: "Applications for courses starting in the 2020 to 2021 academic year are now closed"

config/routes.rb
TobyRet: Good spot - the Rack middleware approach was redirecting all requests to the '/cycle-has-ended' page. This is wrong as we still want to maintain access to certain pages (e.g. T&Cs, Privacy Policy) as well as the healthcheck endpoint. 

Have pushed the changes

TobyRet:
> Any reason we moved away from the content we used last year?
> 
> https://bat-design-history.netlify.app/find-teacher-training/end-of-cycle/
> 
> Mainly I prefer "Courses starting in the academic year",
> 
> So maybe: "Applications for courses starting in the 2020 to 2021 academic year are now closed"

@fofr - Might be worth directing this question to Emma Firth? I think she wrote the copy here - https://trello.com/c/mgCgR4ua/1875-%F0%9F%9A%B2%F0%9F%94%9A-find-landing-page-nearing-eoc-and-closure-notice

lib/rack/display_cycle_has_ended_page.rb
TobyRet: This middleware has been removed now

TobyRet:
> Do we need to redirect all of the pages? Perhaps we need the ping/healtcheck/sitemap pages, as well as the routes to the error controller. Perhaps we can scope this to only the search functionality.

👍 See https://github.com/DFE-Digital/find-teacher-training/pull/412#discussion_r481242912

EmmaFrith:
@fofr this page came to my attention after working on other end of cycle content, where we'd already started using 'this academic year (date)/ next academic year (date)' etc. so was trying to be consistent with other EoC comms. 

other changes include:

- not using / between date ranges, and using 'to' instead, as per gov.uk style
- adding an extra subheading so users can see more quickly at a glance what's going on - and to differentiate between what's going on this year versus next year 
- using 'find' rather than 'view' to continue the "brand" name - also pet peeve - 'view' implies looking and not everyone uses eyes to find things 

This was a v quick content exercise so by no means perfect - but I don't think it's worth changing now!

tijmenb (approved):
👍

PR #413, opened 2020-08-26
use /ping as statuscake healthcheck url

vigneshmsft:

configure the `/ping` as the health check endpoint in statuscake.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
~- [ ] Product review~

### DIFF
--- a/terraform/terraform_prod.tfvars
+++ b/terraform/terraform_prod.tfvars
@@ -1,9 +1,9 @@
 alerts =  {
 ftt = {
     website_name = "prod-find-postgraduate-teacher-training"
-    website_url   = "https://www.find-postgraduate-teacher-training.service.gov.uk/healthcheck"
+    website_url   = "https://www.find-postgraduate-teacher-training.service.gov.uk/ping"
     test_type     = "HTTP"
-    check_rate    = 300
+    check_rate    = 60
     contact_group = [151103]
     trigger_rate  = 0
     custom_header = "{\"Content-Type\": \"application/x-www-form-urlencoded\"}"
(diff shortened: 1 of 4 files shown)


TobyRet (approved):

terraform/terraform_prod.tfvars
```diff
@@ -1,7 +1,7 @@
 alerts =  {
 ftt = {
     website_name = "prod-find-postgraduate-teacher-training"
-    website_url   = "https://www.find-postgraduate-teacher-training.service.gov.uk/healthcheck"
+    website_url   = "https://www.find-postgraduate-teacher-training.service.gov.uk/ping"
     test_type     = "HTTP"
     check_rate    = 300
```
saliceti, line 6: Please reduce interval to 1min

.github/workflows/terraform.yml
```diff
@@ -28,6 +28,11 @@ jobs:
     - name: Checkout the repository to the GitHub Actions runner
       uses: actions/checkout@v2
 
+    - name: Use Terraform 0.12.29
```
saliceti, line 31: This is not required in this PR as statuscake is not currently delivered via github actions. We should create a separate card and PR for that.

saliceti:
Could you please also add 1 more confirmation by default to avoid false positives?
We added it in teaching vacancies: https://github.com/DFE-Digital/teacher-vacancy-service/blob/master/terraform/app/modules/statuscake/statuscake.tf#L11

saliceti (approved):
OK

PR #419, opened 2020-09-01
Update PULL_REQUEST_TEMPLATE.md

tijmenb:
We should add the Trello ticket to PRs.


TobyRet (approved):
We tried to automate this process by activating the Github powerup on Trello (as we used to do on the Find / Publish team) but both Theo and I ran into some issues making it work (specifically, configuring the repo). Doing it manually for the meantime makes sense.

PR #422, opened 2020-09-02
Update Azure template to include end of cycle feature flags

TobyRet:
### Context
A number of changes will take place on Find as part of the end of cycle process. To prepare for this we want to be able to toggle end of cycle functionality easily.

- The azure template has been updated so that we can toggle end of cycle functionality via the Azure portal.
- This will be a lot quicker than raising a PR every time we want to update a feature flag.

### Trello
https://trello.com/c/pF2NLAY9/2070-add-azure-config-for-end-of-cycle-feature-flags

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -144,6 +144,28 @@
       "metadata": {
         "description": "API Key for Google Geocode"
       }
+    },
+    "settingsCycleEndingSoon": {
+      "type": "bool",
+      "defaultValue": false,
+      "metadata": {
+        "description": "When true, requests to '/' will be redirected to '/cycle-ending-soon"
+      }
+    },
+    "settingsCycleHasEnded": {
+      "type": "bool",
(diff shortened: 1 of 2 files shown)


azure/template.json
```diff
@@ -144,6 +144,25 @@
       "metadata": {
         "description": "API Key for Google Geocode"
       }
+    },
+    "settingsCycleEndingSoon": {
+      "type": "string",
```
vigneshmsft, line 149:
```suggestion
      "type": "bool",
      "defaultValue": false
```
could do the same (`type: bool` and `defaultValue: false`) for other parameters too?

TobyRet: Done! 👍

vigneshmsft (approved):

PR #424, opened 2020-09-03
End of cycle - remove 'apply for this course' button

TobyRet:
### Context
- As part of the end of cycle process we do not want users to be able to apply for courses from Sep 18th.

- This change allows the button to be toggle on / off via Azure or local settings (default is set to true / display the button).

### Guidance to review
- To test locally, spin up the app and set the value of `display_apply_button` in your `development.local.yml` to false. When you navigate to a course listing the 'apply for this course' button should not be visible.

### Trello card
https://trello.com/c/vQRmFS25/2067-find-dev-apply-for-this-course-button-changes-18-9-13-10

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/views/courses/_apply.html.erb
+++ b/app/views/courses/_apply.html.erb
@@ -6,12 +6,14 @@
   <h2 class="govuk-heading-l" id="section-apply">Apply</h2>
 
   <% if @course.has_vacancies? %>
-    <p class="govuk-body">
-      <%= link_to "Apply for this course", apply_path(provider_code: course.provider.provider_code, course_code: course.course_code),
-                  class: "govuk-button govuk-button--start",
-                  data: { qa: 'course__apply_link' },
-                  rel: "nofollow" %>
-    </p>
+    <% if Settings.display_apply_button %>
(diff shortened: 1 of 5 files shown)


tvararu (approved):
Looks okay! This is a different approach from what we do in Apply, which is make changes like this automatic based on a timetable and overridable via feature flags (soon: the database) for manual testing. I think for now we can leave this as-is, but it would be good to automate what can be automated in the future and align the two projects in how they handle this type of feature.

TobyRet:
> Looks okay! This is a different approach from what we do in Apply, which is make changes like this automatic based on a timetable and overridable via feature flags (soon: the database) for manual testing. I think for now we can leave this as-is, but it would be good to automate what can be automated in the future and align the two projects in how they handle this type of feature.

Yes agree it would be good to align approaches. Find has no db right now but if it were to merge into the Apply super monolith  this can change. Spoke to Vignesh whilst you were gone and agreed the Azure env var approach was a pragmatic way forward for now.

PR #436, opened 2020-09-09
Fix validations on start wizard when nearing end of cycle

TobyRet:
### Context
- When `Settings.cycle_ending_soon` is set to `true` the LocationsController was not redirecting back to the ‘start/location’ page when validation errors occur. Instead it was redirecting to the ‘/cycle-ending-soon’ page. This is also causing Find end to end tests to fail ([see related PR](https://github.com/DFE-Digital/find-teacher-training-tests/pull/52)).

- New logic has been added to the LocationsController to detect whether `Settings.cycle_ending_soon` is `true` and if so, will redirect back to ‘/start/location’ when a search validation error occurs.

### Before
before

### After 
after

### Trello card
https://trello.com/c/AidwYrAz/2119-fix-broken-find-end-to-end-tests

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -73,7 +73,12 @@ module ResultFilters
     end
 
     def back_to_current_page_if_error(form_params)
-      if flash[:start_wizard]
+      if flash[:start_wizard] && Settings.cycle_ending_soon
+        # In this scenario we do not want to redirect to 'root_path'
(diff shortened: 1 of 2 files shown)


davidgisbey (approved):
LGTM. Is it worth testing the error is showing on the desired page?

PR #450, opened 2020-09-28
terraform modules for PaaS and GitHub action steps to deploy

vigneshmsft:
### Context
Deploy app to QA space.

Terraform files for deploying app to PaaS
GitHub actions steps to run deploy built docker image to QA (PaaS) using terraform.

### Trello card
https://trello.com/c/ba0DQeqX/2207-refactor-terraform-code-for-deploying-to-govuk-paas

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product review


terraform/workspace_variables/qa.tfvars
```diff
@@ -0,0 +1,13 @@
+# PaaS
+paas_sso_code                      = ""
+paas_app_environment               = "qa"
+paas_rails_env                     = "qa"
+paas_space_name                    = "bat-qa"
```
saliceti, line 5: This could be calculated in the module so we don't have to pass it through

terraform/workspace_variables/qa.tfvars
```diff
@@ -0,0 +1,13 @@
...
+paas_app_start_timeout             = "180"
+paas_web_app_deployment_strategy   = "blue-green-v2"
```
saliceti, line 7: We will always use blue-green-v2 for the webapp so I suggest to remove the variable and hardcode  blue-green-v2  in the module

terraform/workspace_variables/qa.tfvars
```diff
@@ -0,0 +1,13 @@
+# PaaS
+paas_sso_code                      = ""
+paas_app_environment               = "qa"
+paas_rails_env                     = "qa"
+paas_space_name                    = "bat-qa"
+paas_app_start_timeout             = "180"
```
saliceti, line 6: This should be the same for all environments. We could also hardcode it in the module.

saliceti (approved):
OK

PR #456, opened 2020-10-01
Fix broken specs resulting from change in cycle

TobyRet:
### Context
Many of the api test stubs are hard coded to the `2020` cycle year. When stubbing the API we need to concatenate `Settings.current_cycle` into API URL stings so that when the current cycle is updated from `2020` to `2021`, the tests do not break.

- Create [course_url helper](https://github.com/DFE-Digital/find-teacher-training/pull/456/files#diff-3bb5ea059ab3c1c6dbccd241014d50b2) so that it is only defined in one place. 
- Concatenate the settings values for `cycle_year` and `teacher_training_api.base_url` so that any changes in these settings will not break specs

### Guidance to review
- Lots of file changes sorry, but they're small changes 
- Potentially there is scope to do some more url refactoring but will card this up as priority is to get the specs working

### Trello card
https://trello.com/c/CcJuw6ym/2254-new-cycle-2021-fix-broken-find-tests

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/spec/controllers/result_filters/subject_controller_spec.rb
+++ b/spec/controllers/result_filters/subject_controller_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe ResultFilters::SubjectController do
   describe "#new" do
     context "when subjects is a comma delimited string" do
       it "returns a 200" do
-        stub_request(:get, "http://localhost:3001/api/v3/subject_areas?include=subjects")
+        stub_request(:get, "#{Settings.teacher_training_api.base_url}/api/v3/subject_areas?include=subjects")
            .with(
              headers: {
                "Accept" => "application/vnd.api+json",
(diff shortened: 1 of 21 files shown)


AgaDufrat (approved):
LGTM. 

I just noticed that in the specs which use api response fixtures (eg suggested searches) we stub request with `Settings.current_cycle` in the URL but the data still has "recruitment_cycle_year" hardcoded. Just a thought, it's not going to break anything and I don't think it's worth changing.

TobyRet:
> se api response fixtures

Good spot @AgaDufrat 
I'm going to create a separate ticket to re-generate the fixtures (or figure out a way to update the recruitment cycle year dynamically)
Thanks for the review!

PR #473, opened 2020-10-07
Throttle autocomplete calls

tvararu:
### Context

Our autocomplete hammers our server with too many requests.

Use `lodash.throttle` which supports both leading and trailing calls.

### Guidance to review

Test it in the review app, check the dev console.

### Trello card

https://trello.com/c/FXy5da3M/2299-add-js-autocomplete-back-to-find-provider-and-location-search

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -21,16 +21,16 @@ new backLink($backLink).init();
 const $toggle = document.querySelector('[data-module="toggle"]');
 new toggle($toggle).init();
 
-// initAutocomplete({
-//   element: "location-autocomplete",
-//   input: "location",
-//   path: "/location-suggestions",
-// });
-// initAutocomplete({
-//   element: "provider-autocomplete",
-//   input: "provider",
-//   path: "/provider-suggestions",
-// });
+initAutocomplete({
+  element: "location-autocomplete",
+  input: "location",
+  path: "/location-suggestions",
+});
+initAutocomplete({
+  element: "provider-autocomplete",
+  input: "provider",
+  path: "/provider-suggestions",
+});
 
 loadAnalytics();
 new initCookieBanner();
--- a/app/webpacker/scripts/autocomplete.js
+++ b/app/webpacker/scripts/autocomplete.js
@@ -1,4 +1,5 @@
 import accessibleAutocomplete from "accessible-autocomplete";
+import throttle from 'lodash.throttle'
 
 export const getPath = (endpoint,query) => {
   return `${endpoint}?query=${query}`;
@@ -47,7 +48,7 @@ const initAutocomplete = ({element, input, path}) => {
         name: $input.name,
         defaultValue: $input.value,
         minLength: 3,
-        source: request(path),
+        source: throttle(request(path), 500),
         templates: {
           inputValue: inputValueTemplate,
           suggestion: suggestionTemplate


TobyRet (commented):
LGTM

TobyRet (approved):

tijmenb:
I'm not 100% sure this is enough. For a test, I typed "gloucestershire" fairly slowly into the search box. It still results in a fair number of requests:

Screenshot 2020-10-08 at 10 23 42

Also, it seems the dropped requests are cancelled on the client side - Heroku still sees them:

```
2020-10-08T09:23:24.268323+00:00 heroku[router]: at=info method=GET path="/provider-suggestions?query=glouce" host=bat-find-pr-473.herokuapp.com request_id=066fcd7c-4a10-447e-beea-bc6ae4754016 fwd="51.7.198.26" dyno=web.1 connect=0ms service=267ms status=200 bytes=699 protocol=https

2020-10-08T09:23:26.539084+00:00 heroku[router]: sock=client at=warning code=H27 desc="Client Request Interrupted" method=GET path="/provider-suggestions?query=glouces" host=bat-find-pr-473.herokuapp.com request_id=deca89d4-4193-4559-ace6-8fccf0560e76 fwd="51.7.198.26" dyno=web.1 connect=0ms service=785ms status=499 bytes= protocol=https

2020-10-08T09:23:26.761852+00:00 heroku[router]: sock=client at=warning code=H27 desc="Client Request Interrupted" method=GET path="/provider-suggestions?query=gloucest" host=bat-find-pr-473.herokuapp.com request_id=143ac83e-c9e3-4dcd-81e2-262b6cb4a364 fwd="51.7.198.26" dyno=web.1 connect=0ms service=658ms status=499 bytes= protocol=https

2020-10-08T09:23:27.009408+00:00 heroku[router]: at=info method=GET path="/provider-suggestions?query=gloucester" host=bat-find-pr-473.herokuapp.com request_id=c9259c2c-af51-41dd-b16f-5311414bbd29 fwd="51.7.198.26" dyno=web.1 connect=0ms service=393ms status=200 bytes=699 protocol=https
```

I'm a bit unsure whether this means the request has reached Rails, and is therefore still executing in the background (just not returning anything). Not sue how it works on Azure as well.

So couple of questions for us to answer:

- Are there patterns for this other than throttling? Like waiting 1s after last keypress to start autocomplete?
- If we cancel requests, can/does Rails actually stop executing? 
- Can we track the number of requests per "session", so we know what the effect of this is?

tvararu:
> Are there patterns for this other than throttling? Like waiting 1s after last keypress to start autocomplete?

That would be debouncing; when I tried it, it felt quite annoying, because the autocomplete says "No results found" while you wait. We could disable that message completely. I'll give that a push so we can try it out.

> If we cancel requests, can/does Rails actually stop executing?

I think it keeps going and then it will try to reply to the previous requests. Since the connection has been closed, that should no-op. But the effort is already spent in trying to fetch the results for that request. 

> Can we track the number of requests per "session", so we know what the effect of this is?

Do you mean like setting up a performance dashboard where we can track the number of requests over time to the autocomplete endpoint, so we can estimate what if any effect this has? We would need to build that, but we could base it off of the existing logs.

tvararu:
@tijmenb I've pushed up an alternative implementation which uses a 1 second debounce and does not show the `No results found` message, WDYT?

tvararu:
Discussed further with @fofr on Slack. The debounced 1s approach does not seem ideal. It makes the autocomplete feel inert and unresponsive.

I strongly feel that throttling by 300ms is better. According to the internet(tm), the average typist has a typing speed of ~40 words per minute, or 3 characters per second, 1 every ~300ms. I don't think we should aim to throttle/debounce for users that type very slowly, as that would make the experience worse for the majority.

In the interest of getting this useful functionality back up sooner than later, shall we go with the initial approach combined with some better metrics to track if it's actually saving our server from unnecessary load? I can also look into whether it's possible to cancel a request in Rails if the requester has hung up.

TobyRet:
@tvararu @tijmenb 
Would a happy medium for now be to up the throttle to 500ms? I agree - 300ms is possibly a bit too low in the instance where someone is reasonably fast at typing.

tvararu:
@TobyRet pushed

PR #529, opened 2020-11-16
Send a Sentry error when an error is returned in the body of the gcp response

davidgisbey:
### Context

At the moment, when the locations autocomplete fails, it fails silently. This is because GCP returns errors in the body (see image).

image

- Check the body for an error_message. If one is present, then send a Sentry error with the GCP error_message.

### Trello card

https://trello.com/c/fH1nK2ko/2498-log-sentry-error-when-autocomplete-fails

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product review

### DIFF
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -9,10 +9,12 @@ class LocationSuggestion
 
       response = get("#{Settings.google.places_api_path}?#{query.to_query}")
 
-      if response.success?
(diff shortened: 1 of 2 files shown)


george-bit (approved):
LGTM 👍

TobyRet (approved):

PR #548, opened 2020-12-07
Move CSS into component partials

frankieroberto:
This moves some CSS which was in `application.scss` into separate partials, named after their components.

`govuk-list--dash` is renamed to `app-list--dash` to follow the [Design System convention](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/) that the `govuk-` prefix is changed to `app-` for app-specific modifiers.


paulrobertlloyd (approved):
Lots of improvements we can make around organising CSS, but this is a great start. 👍

PR #571, opened 2020-12-17
WIP: Browser back button bug

TobyRet:
### Context
Please see Trello card with description and instructions for replicating the bug: https://trello.com/c/GG6ZLJAr/1979-dev-going-back-to-location-filter-on-results-page

### Findings

When the user clicks the back button all the way back to the start of the search, the `flash[:start_wizard]` param is not being set in the [LocationsController](https://github.com/DFE-Digital/find-teacher-training/blob/master/app/controllers/result_filters/location_controller.rb#L9) as a hard refresh does not occur. As a result, a subsequent search [redirects straight to the results page](https://github.com/DFE-Digital/find-teacher-training/blob/master/app/controllers/result_filters/location_controller.rb#L71) rather than via the subjects form. 

### Hacking a solution

You will see in the changes files it's possible to detect whether a user has arrived at a given page by checking `performance.navigation.type`. [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation)

Whilst this does fix the issue, as the page loads the radio button is briefly selected before becoming deselected.

### Questions

Part of me thinks that this is not a major bug but if it considered important, we will need a better solution than the one I have proposed.

It's possible also to see if delaying fixing is an option, in the event @paulrobertlloyd wishes to make significant changes to the app and  it's flows.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -10,9 +10,11 @@ import initAutocomplete from "scripts/autocomplete";
 import toggle from "scripts/toggle";
 import { loadAnalytics } from "scripts/analytics.js";
 import initCookieBanner from "scripts/cookie-banner";
+import refreshStartPage from "scripts/refresh-start-page";
 
 initAll();
 
+refreshStartPage();
 window.initLocationsMap = initLocationsMap;
 
 const $backLink = document.querySelector('[data-module="back-link"]');
--- a/app/webpacker/scripts/refresh-start-page.js
+++ b/app/webpacker/scripts/refresh-start-page.js
@@ -1,0 +1,6 @@
+const refreshStartPage = () => {
+    if (window.location.pathname === '/' && performance.navigation.type === 2) {
+        location.reload(true);
+    }
+}
+export default refreshStartPage;


frankieroberto:
I'm not sure this is the right fix – I think you'd expect, if going all the way back to the first screen, for it to remember the location you previously added (text box pe-filled), and then when pressing continue for it to go to the subject selection screen, but for your previous answers to still be pre-checked.

I'm not sure this is a significant enough bug to fix just yet though, and we'll soon be re-examining these flows anyway.

TobyRet:
Another option could be to remove the need for storing state in flash and finding an alternate way to redirect to the subjects form - perhaps through params? 🤔

frankieroberto:
@TobyRet yeah, I think storing all the params in the query string is the way to go. That wouldn't work when pressing the browser back button, but the browser _should_ restore the state for itself.  I think at the moment the location autocomplete interferes with this though, which is arguably a bug with that component... 🤔

TobyRet:
Closing - going to wait until Find rebuild is done as it may no longer be an issue

PR #574, opened 2020-12-21
Change default search results (vacancies)

TobyRet:
### Context
```
As a… user of Find
I need to... see all course with and without vacancies
So that... I can see the full range of courses and providers available in my area
```

- When searching we now display courses with or without vacancies by default
- Always display whether a course has vacancies or not in the course results summary

### Guidance to review
- Run any search and you will see the search results include courses with and without vacancies
- Regardless of whether you run a search or filter the search by 'vacancies' each course results summary will display a row indicating whether it has vacancies or not

### Trello card
https://trello.com/c/LBpLaXJ2/2693-%F0%9F%97%BA-dev-start-by-showing-all-course-with-and-without-vacancies-by-default

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product review

### DIFF
--- a/app/views/results/index.html.erb
+++ b/app/views/results/index.html.erb
@@ -135,12 +135,10 @@
                 <%= smart_quotes(course.provider.decorate.short_address) %>
               </dd>
             <% end %>
-            <% if @results_view.vacancy_filter?  %>
-              <dt class="govuk-list--description__label">Vacancies</dt>
-              <dd data-qa="course__has_vacancies">
-                <%= course.decorate.has_vacancies? %>
-              </dd>
-            <% end %>
+            <dt class="govuk-list--description__label">Vacancies</dt>
+            <dd data-qa="course__has_vacancies">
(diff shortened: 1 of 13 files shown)


app/views/results/index.html.erb
```diff
@@ -135,12 +135,10 @@
                 <%= smart_quotes(course.provider.decorate.short_address) %>
               </dd>
             <% end %>
-            <% if @results_view.vacancy_filter?  %>
-              <dt class="govuk-list--description__label">Vacancies</dt>
-              <dd data-qa="course__has_vacancies">
-                <%= course.decorate.has_vacancies? %>
-              </dd>
-            <% end %>
+            <dt class="govuk-list--description__label">Vacancies</dt>
```
TobyRet, line 138: We now indicate whether a course has vacancies or not, regardless of the vacancies filter value

davidgisbey (approved):
lgtm

PR #592, opened 2021-01-04
Update error message wording

frankieroberto:
@gregknight1 reviewed the error message text across the service, to make them more usable and consistent with the GOV.UK Design System.

### Bonus fix

This also changes the design of the errors within conditionally-revealed fields, to place the red line next to the revealed field only, rather than the whole of the radio fieldset.

#### Before

Screenshot 2021-01-04 at 13 49 39

#### After

Screenshot 2021-01-04 at 15 56 24

### Trello card

https://trello.com/c/i8qZkt3C/2818-improve-error-messages-in-find
https://trello.com/c/QitTcgV3/2764-dac-quick-wins

### DIFF
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -19,11 +19,11 @@
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
           <h1 class="govuk-fieldset__heading">Find courses by location or by training provider</h1>
         </legend>
-        <div class="govuk-form-group<%= " govuk-form-group--error" if flash[:error] %>" id="search-options">
+        <div class="govuk-form-group" id="search-options">
           <% if flash[:error] and location_error? === false and provider_error? === false%>
             <p class="govuk-error-message">
               <% if flash[:error].kind_of?(Array) %>
-                <%= flash[:error].first %>
+                <%= flash[:error].last %>
               <% else %>
                 <%= flash[:error] %>
               <% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -49,20 +49,22 @@ en:
     pgde_with_qts: "PGDE with QTS"
   location_filter:
     errors:
-      no_option: "Please choose an option"
-      unknown_location: "We couldn’t find this location, please check your input and try again"
-      missing_location: "Please enter a postcode, city or town in England"
-      missing_provider: "We couldn’t find this provider, please check your information and try again"
-      blank_provider: "You need to add some information"
-      invalid_provider: "Please enter a minimum of two characters"
+      no_option: Select an option to find courses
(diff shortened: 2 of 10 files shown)


config/locales/en.yml
```diff
@@ -49,20 +49,22 @@ en:
     pgde_with_qts: "PGDE with QTS"
   location_filter:
     errors:
-      no_option: "Please choose an option"
-      unknown_location: "We couldn’t find this location, please check your input and try again"
-      missing_location: "Please enter a postcode, city or town in England"
-      missing_provider: "We couldn’t find this provider, please check your information and try again"
-      blank_provider: "You need to add some information"
-      invalid_provider: "Please enter a minimum of two characters"
```
frankieroberto, line 57: Note: we agreed that "Please enter a minimum of two characters" could be dropped, as the "Enter a real school, university or training provider" message also applies here, so we don't need the separate `invalid_provider` key.

app/views/result_filters/location/_form.html.erb
```diff
@@ -19,11 +19,11 @@
         <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
           <h1 class="govuk-fieldset__heading">Find courses by location or by training provider</h1>
         </legend>
-        <div class="govuk-form-group<%= " govuk-form-group--error" if flash[:error] %>" id="search-options">
+        <div class="govuk-form-group" id="search-options">
           <% if flash[:error] and location_error? === false and provider_error? === false%>
             <p class="govuk-error-message">
               <% if flash[:error].kind_of?(Array) %>
-                <%= flash[:error].first %>
+                <%= flash[:error].last %>
```
frankieroberto, line 26: This fixes a bug where the error message in the summary and next to the field is currently sometimes different. This happens when the error message applies to a conditionally-revealed field, where the code uses an array of two items to identify which field was in error, with the first item in the array being the label of the field, and the second item being the actual error message. So the last item of the array should be shown instead of the first one.

See example of the issue:


Screenshot 2021-01-04 at 15 52 05

An array within the flash isn't the ideal way to store these errors, so we should come back and fix this as part of a wider refactor.

config/locales/en.yml
```diff
@@ -49,20 +49,22 @@ en:
...
   subject_filter:
     errors:
-      no_option: "Please choose at least one subject"
+      no_option: Select at least one subject
+  qualification_filter:
+    errors:
+      no_option: Select at least one qualification
   cookie_preferences:
-    no_option_error: "Please choose an option"
+    no_option_error: Select yes if you accept Google Analytics cookies
```
frankieroberto, line 67: @gregknight1 spotted another one here. Let me know if this is ok, or if you have any better suggestions?

gregknight1: @frankieroberto That's fine - thanks.

paulrobertlloyd (approved):

PR #602, opened 2021-01-13
Replace Rubocop config with link to Apply

stevehook:
### Context

We want to use the same Rubocop config across Find and Apply. This PR attempts to make it easier to keep the two in sync by referencing the Apply config directly from its public URL on Github.

- [x] Update `.rubocop.yml` to link to the Apply `.rubocop.yml` rather than contain a copy-and-pasted snapshot.
- [x] Also updates `.gitignore` to ignore cached downloaded config files that Rubocop leaves in `RAILS_ROOT`.

### Guidance to review

- Probably best to just give it a try locally. Does Rubocop work as expected after this change? (I tested locally and Rubocop passed with the current code and failed when I added a violation.)
- The main risk here is that the Apply config file moves. Rubocop throws a fairly clear error if the URL in the config points to a missing.

### Trello card

https://trello.com/c/Jc8TqR8u/2825-dev-use-shared-rubocop-config-across-find-and-apply

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


davidgisbey (approved):
Awesome. It's working nicely locally. I think the Apply file moving isn't a huge deal. It's unlikely that it will ever move and like you said the error message is clear.

PR #605, opened 2021-01-14
Don't pass in the 'has_vacancies' filter param to the API when requesting courses with OR without vacancies

TobyRet:
### Context

Currently when we request courses with or without vacancies (now the default behaviour on Find) we are passing in the filter param `{ has_vacancies: true}`

This is incorrect from in terms of the Teacher Training API specification - we should just not pass in anything at all as the default behaviour is to return all courses, regardless of whether they have vacancies or not. See https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#schema-coursesingleresponse

- Add the filter param `{has_vacancies: true}` only when we want courses with vacancies. When we want courses with or without vacancies, don't pass a filter param. Currently we do no request only courses without vacancies, but if we were to then we would have to add the filter param `{ has_vacancies: false }`

### Guidance to review
Make sense?

### Trello card
https://trello.com/c/5etHxUms/2856-dont-pass-in-the-hasvacancies-filter-param-to-the-api-when-requesting-courses-with-or-without-vacancies

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


JR-G (approved):

davidgisbey (approved):
Yea that makes sense to me Toby. Defaults to all courses and we explicitly apply the filter when needed.

PR #609, opened 2021-01-15
Reinstate project specific Rubocop rules

stevehook:
### Context

These were mistakenly overwritten when changing the way we share rules
between Apply and Find in
https://github.com/DFE-Digital/find-teacher-training/pull/602

Just re-adds the lost rules

### Guidance to review

Does this make sense? Have I missed anything still?

### Trello card

https://trello.com/c/Jc8TqR8u/2825-dev-use-shared-rubocop-config-across-find-and-apply

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,2 +1,26 @@
 inherit_from:
   - https://raw.githubusercontent.com/DFE-Digital/apply-for-teacher-training/master/.rubocop.yml
+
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
Yea, i should have picked this up at review before. Sorry!

PR #611, opened 2021-01-15
Refactor stubbed requests that use fixtures

malcolmbaig:
### Context

We stub various requests across the test suite, returning the contents of one of several JSON fixture files. This tends to be quite verbose, so this PR adds some helpers to take care of this stubbing.

- A couple of test suite improvements - support for `--only-failures` and fixing some deprecation warnings.
- Replace every inline case of a request being stubbed to return a JSON fixture with a helper that does the same.

### Guidance to review
- Per-commit highly recommended, each set of replacements has been batched by the kind of endpoint being stubbed.
- Is it clear enough? Has anything been missed?
 
### Trello card
https://trello.com/c/cAWS3e8v
### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/features/cookie_banner_spec.rb
+++ b/spec/features/cookie_banner_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe 'cookie banner', type: :feature do
+  include StubbedRequests::Courses
+
   let(:results_page) { PageObjects::Page::Results.new }
   let(:params) {}
   let(:subject_areas) do
@@ -22,12 +24,7 @@ describe 'cookie banner', type: :feature do
   let(:base_parameters) { results_page_parameters }
 
   def stub_results_request
-    stub_request(:get, courses_url)
-      .with(query: base_parameters)
-      .to_return(
-        body: File.new('spec/fixtures/api_responses/ten_courses.json'),
-        headers: { "Content-Type": 'application/vnd.api+json; charset=utf-8' },
-      )
+    stub_courses(query: base_parameters, course_count: 10)
   end
 
   before do
--- a/spec/features/cookie_banner_spec.rb
+++ b/spec/features/cookie_banner_spec.rb
@@ -2,6 +2,7 @@ require 'rails_helper'
 
 describe 'cookie banner', type: :feature do
   include StubbedRequests::Courses
+  include StubbedRequests::Subjects
 
   let(:results_page) { PageObjects::Page::Results.new }
   let(:params) {}
@@ -29,8 +30,7 @@ describe 'cookie banner', type: :feature do
 
   before do
     stub_results_request
-    stub_subjects_request
-
+    stub_subjects
     visit results_path(params)
   end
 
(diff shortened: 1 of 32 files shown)


spec/spec_helper.rb
```diff
@@ -51,7 +51,7 @@
   #   # Allows RSpec to persist some state between runs in order to support
   #   # the `--only-failures` and `--next-failure` CLI options. We recommend
   #   # you configure your source control system to ignore this file.
-  #   config.example_status_persistence_file_path = "spec/examples.txt"
+  config.example_status_persistence_file_path = 'tmp/rspec-failures'
```
davidgisbey, line 54: 👌

davidgisbey:
I'll go through this properly, but it looks great at a first look 👍

spec/features/cookie_banner_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 describe 'cookie banner', type: :feature do
+  include StubbedRequests::Courses
```
TobyRet, line 4: Probably this comes down to whether the team prefers implicit  vs explicit declaration but one option would be to include the helper via the `spec_helper` so that you do not need to add a list of `includes` in lots of specs. 

```
# spec_helper.rb
RSpec.configure do |c|
  c.include StubbedRequests::Courses
  c.include StubbedRequests::Providers
  # etc
end
```

https://relishapp.com/rspec/rspec-core/v/3-8/docs/helper-methods/define-helper-methods-in-a-module

TobyRet (commented):
Looking good! Much tidier.
Just one small comment / question re: including spec helper modules

spec/support/stubbed_requests/courses.rb
```diff
@@ -0,0 +1,32 @@
...
+        'two_courses.json'
+      when 4
+        'four_courses.json'
+      when 10
+        'ten_courses.json'
+      end
+    end
+
+    def courses_url
+      "#{Settings.teacher_training_api.base_url}/api/v3/recruitment_cycles/#{Settings.current_cycle}/courses"
```
TobyRet, line 29: Not for this PR but wondering whether we extract the API version (v3 ,  public/v1) to Settings 🤔 

That way if we need to change the version again quickly it only needs to be done in one place

spec/features/cookie_banner_spec.rb
davidgisbey: Yea this was one thought I had when reviewing too. Normally I'd be on the side of making it really explicit, but they're pretty consistently used in the feature specs so it might be worth just including in the spec helper.

I think either fine though tbh.

davidgisbey (approved):
Looks good. That's so much tidier and less noisy 👍 

There's the one question Toby brought up, but am happy to approve for whichever approach we take.

spec/features/cookie_banner_spec.rb
malcolmbaig: I quite like the explicit inclusions at the top of the file - they're informative in that they give a quick overview of which requests are being stubbed and the overall degree of stubbing in a given spec file. Perhaps one to refactor away if it becomes too laborious?

TobyRet (approved):

PR #621, opened 2021-01-22
Use footer component from govuk-components gem

paulrobertlloyd:
### Context

After discussion with @peteryates about allowing more customisation of the footer component in the `govuk-components` gem (https://github.com/DFE-Digital/govuk-components/pull/95), this PR previews the resulting change on Find.

Ultimately, because our current footer design deviates from the layouts allowable by the Design System, in effect all that this component does is replace the following containing elements:

```html
<footer class="govuk-footer govuk-!-display-none-print" role="contentinfo">
  <div class="govuk-width-container">
    <div class="govuk-footer__meta">
      …
    </div>
  </div>
</div>
```

This feels like a good compromise. Should the design of the footer align more closely to the design of the Design System component in the future, we can use more of the options provided by this gem.

Happy with this approach?

### Guidance to review

Pending release of the next version of the `govuk-components` gem.

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


stevehook (approved):
:+1: Markup looks good. Testing locally looks identical to the original.

PR #632, opened 2021-02-01
Formalise how Find feature flags work

malcolmbaig:
### Context

We're planning significant feature work on Find, and so need some kind of mechanism for adding and switching feature flags.

There were a few different options identified, as outlined [here](https://ukgovernmentdfe.slack.com/archives/CN1MCQCHZ/p1611845790103500). In summary:

---

**Option 1 - Use the Apply feature flag API**
Pros:
- Can use the existing (secured, audited) feature flag dashboard in Apply to set flags for Find

Cons:
- Probably need to implement some kind of caching of API responses for resilience
- Need to add setup to ensure this works smoothly in development and test

**Option 2 - Build a feature flag mechanism local to Find**
Pros:
- Not dependent on Apply uptime/availability
- Can use an established library to implement and configure (eg - flipper)

Cons:
- Need to add some kind of storage for feature flags (Redis/Postgres)
- Needs a UI to be built (with security and auditing), otherwise flipping features is dependent on developers

**Option 3 - Just use the settings.yml and deploy to flip features**
Pros:
- Minimal changes, just formalise/document existing use of the config gem to handle features

Cons:
- Needs a deploy to flip features - much slower, less accessible to wider team

---

In essence, building functionality comparable to Apply (persistence, a password-protected UI, an audit trail of changes) requires a fair bit of additional complexity to be added to the Find codebase—largely because of the absence of a datastore. Given this, we've decided to proceed with Option 3 as it introduces the least additional complexity at the expense of ease-of-use. If this proves too much of a hassle, we can revisit one of the other options.

- Store all feature flag definitions in settings.yml
- Add a `FeatureFlag` helper class to check flag state - this provides a standard interface which is easily grep'd and can remain unchanged if the underlying implementation changes in future.
- Add a spec helper to standardise activating/deactivating features in tests
- Document the approach

### Guidance to review
- Probably worth reading the doc first for an initial outline before looking at the code changes
- Does the approach make sense? Are there any obvious obstacles that haven't been accounted for?
 
### Trello card
https://trello.com/c/DqMUaSiL

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


malcolmbaig:
Making a quick update to specs

PR #638, opened 2021-02-03
Update entry requirements partial to use standardised content

george-bit:
### Context
As part of a content overhaul on the find application we are updating the copy to
be more in line with the styling on apply.

This PR mirrors the design prototype. https://find-prototy-feb-design-i6fqfe.herokuapp.com/course/B72/X100

### Guidance to review

This is a pretty granular solution as I have had limited success in finding the required
info on the TTAPI as a response. Would be interested if the copy needs to be updated there
as opposed to here.

### Trello card

https://trello.com/c/QGcRTsq4/2937-%F0%9F%97%BA-dev-add-standardised-content-for-qualifications-to-course-detail-pages

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


paulrobertlloyd:
Oops, just to be clear, none of the existing content in this section should be removed (i.e. `course.personal_qualities` and `course.other_requirements`), just that this block of standardised content should appear above this content (should it exist).

paulrobertlloyd:
Also, don’t know if we are using view components on Find, but this is ripe for making into a component. Maybe that’s introduced once we have two structured boxes on a course page?

george-bit:
> Oops, just to be clear, none of the existing content in this section should be removed (i.e. `course.personal_qualities` and `course.other_requirements`), instead this block of standardised content should appear above that content (should it exist).

No worries @paulrobertlloyd, was not sure as was going by how it was showing in the prototype. It's simple enough to add back in and separate into its own partial file. To be absolutely sure would it just be the info within the app advice card or would it include the two bullets that come after it too?

george-bit:
> Also, don’t know if we are using view components on Find, but this is ripe for making into a component. Maybe that’s introduced once we have two structured boxes on a course page?

Oh so you mean in concert with the other sections? So we would create a generic app advice component that we could reuse across the application? 🤔 

I think it could be useful but would have to think about it and discuss a bit with the devs. Would this perhaps be restrictive to how copy could be written? Apologies, if this has already all been discussed and agreed.

paulrobertlloyd:
> To be absolutely sure would it just be the info within the app advice card or would it include the two bullets that come after it too?

Probably best demonstrated by looking at the prototype: https://github.com/DFE-Digital/find-teacher-training-prototype/blob/59ac5ef686f4731973ae26b32f2a34a153a9c005/app/views/course.njk#L93-L116

Apart from the headings, most of content on this page comes directly from the API, so some content may contain bullets, some not.

paulrobertlloyd:
As to a component, I’d imagine something like this:

```erb
<!-- Component HTML -->
<aside class="app-advice">
  <h3 class="app-advice__title"><%= title %></h3>
  <div class="app-advice__body">
    <%= content %>
  </div>
</aside>

<!-- Which could be called something like this -->
<%= render AdviceComponent.new(title: "Qualifications") do %>
  <p class="govuk-body">You should have some</p>
<% end %>
```

george-bit:
@paulrobertlloyd - Changes made inline with suggestions :)

app/views/courses/_entry_requirements_qualifications.html.erb
```diff
@@ -1,7 +1,11 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-entry">Entry requirements</h2>
   <% if course.required_qualifications.present? %>
-    <h3 class="govuk-heading-m">Qualifications</h3>
+    <%= render AdviceComponent.new(title: "Qualifications") do %>
+      <%= render partial: 'courses/english_entry_requirements' %>
```
paulrobertlloyd, line 5: ‘English’ isn’t really relevant:

```suggestion
      <%= render partial: 'courses/entry_requirements' %>
```

paulrobertlloyd:
Just the one smol change, but otherwise, 🚢 !

app/views/courses/_entry_requirements.html.erb
```diff
@@ -0,0 +1,12 @@
+<p class="govuk-body">To be a teacher in England you need:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>a degree or an equivalent qualification</li>
+  <li>English and Maths GCSEs at grade 4 (C) or above</li>
+  <li>For primary teaching, a science subject GCSE at grade 4 (C) or above</li>
+</ul>
+<p class="govuk-body">You can find out more about entry requirements, and how it works if you have studied overseas, at <%= link_to "Get Into Teaching", "https://beta-getintoteaching.education.gov.uk/steps-to-become-a-teacher" %>.</p>
```
davidgisbey, line 7: this needs be gov_uk_link_to please 🙏

app/views/courses/_entry_requirements.html.erb
```diff
@@ -0,0 +1,12 @@
...
+
+<h4 class="govuk-heading-s">Refreshing your subject knowledge</h4>
+<p class="govuk-body">If you have the right qualities and qualifications to teach but need to refresh your subject knowledge, you might be asked to complete a <%= link_to "subject knowledge enhancement (SKE) course", "https://getintoteaching.education.gov.uk/explore-my-options/teacher-training-routes/subject-knowledge-enhancement-ske-courses" %>.</p>
```
davidgisbey, line 10: as does this

davidgisbey (approved):
Looking good. A couple of links just need the to be govuk_link_to's

app/views/courses/_entry_requirements.html.erb
george-bit: I've done so for both, previously the link style was being inherited from the class on the <p> tag. Is this a govuk standard for screen readers or something? Just asking so I know for next time 🙃

paulrobertlloyd: `govuk_link_to` is an extension of the Rails link helper which adds the `govuk-link` class to the link. You need to add this class to links so they use the right colour (a CSS architecture decision which causes no end of problems, but that’s a whole other rant 😅)

PR #639, opened 2021-02-04
Update school placement partials to use standardised content

JR-G:
### Context

In order to improve guidance for candidates trying to understand
how school placements work, standardised content has been added.

As there is a slight difference in the content shown for University
courses and SCITT/School Direct courses, two partials have been added.
|SCITT/School Direct|University|
|-----|-----|
|SCITT|University|

### Trello card

https://trello.com/c/GC2b4xrH/2936-%F0%9F%97%BA-dev-add-standardised-content-for-where-you-will-train-to-course-detail-pages

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -37,4 +37,10 @@ class Course < Base
   def further_education?
     level == 'further_education' && subjects.any? { |s| s.subject_name == 'Further education' || s.subject_code = '41' }
   end
+
+  def travel_to_work_areas
+    travel_to_work_areas = site_statuses.map(&:site).map { |site| site.london_borough || site.travel_to_work_area }.uniq
+
+    travel_to_work_areas.to_sentence(last_word_connector: ' and ')
+  end
 end
--- a/app/views/courses/_about_schools.html.erb
+++ b/app/views/courses/_about_schools.html.erb
@@ -1,6 +1,13 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-schools"><%= course.placements_heading %></h2>
   <div data-qa="course__about_schools">
+      <%= render AdviceComponent.new(title: "Where you will train") do %>
+        <% if course.provider_type == "university" %>
+          <%= render partial: 'courses/placement/hei', locals: { course: course }  %>
+        <% else %>
+          <%= render partial: 'courses/placement/scitt', locals: { course: course }  %>
+        <% end %>
+      <% end %>
     <%= markdown(course.how_school_placements_work) %>
   </div>
 </div>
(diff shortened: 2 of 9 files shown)


app/models/course.rb
```diff
@@ -37,4 +37,16 @@ def university_based?
...
+
+  def travel_to_work_areas
+    travel_to_work_areas = sites.map { |site| site.london_borough || site.travel_to_work_area }.uniq
+    return if travel_to_work_areas.blank?
+
+    if travel_to_work_areas.length == 1
+      travel_to_work_areas.first
+    else
+      last_travel_to_work_area = travel_to_work_areas.pop
+      travel_to_work_areas.join(', ') + " and #{last_travel_to_work_area}"
```
george-bit, line 49: When you pop the last value off the array it should return as a string so you don't need to interpolate, wrote mine like this in the end.
```suggestion
      travel_to_work_areas.join(', ') + " and " + last_travel_to_work_area
```

app/views/courses/_about_schools.html.erb
```diff
@@ -1,6 +1,13 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-schools"><%= course.placements_heading %></h2>
   <div data-qa="course__about_schools">
+      <%= render AdviceComponent.new(title: "Where you will train") do %>
+        <% if course.provider_type == "university" %>
```
george-bit, line 5: I think this should be
```suggestion
        <% if course.provider_type == "HEI" %>
```

davidgisbey: So the enum value is 
image

app/models/course.rb
```diff
@@ -37,4 +37,16 @@ def university_based?
...
+    end
+  end
```
davidgisbey, line 51: Sorry i know this was my idea, but i was thinking there must be a method that does this. It turns out you can just use .to_sentence

image

It also behaves with an empty array 👌 

image

davidgisbey: Just realised that it leaves the last comma in as default. You can get around that like this.

```
["Brighton", "England", "Wales"].to_sentence
"Brighton, England, and Wales"

["Brighton", "England", "Wales"].to_sentence(last_word_connector: ' and ')
"Brighton, England and Wales"
```

app/controllers/courses_controller.rb
```diff
@@ -7,6 +7,7 @@ def show
     @course = Course
       .includes(:subjects)
       .includes(site_statuses: [:site])
+      .includes(:sites)
```
davidgisbey, line 10: It's not a huge deal but i think we should be accessing sites through site_statuses. Otherwise, we're getting sites data back twice

app/models/course.rb
```diff
@@ -37,4 +37,16 @@ def university_based?
   def further_education?
     level == 'further_education' && subjects.any? { |s| s.subject_name == 'Further education' || s.subject_code = '41' }
   end
+
+  def travel_to_work_areas
+    travel_to_work_areas = sites.map { |site| site.london_borough || site.travel_to_work_area }.uniq
```
davidgisbey, line 42: so here we would do site_statuses.map(&site).map { |site| site.london_borough || site.travel_to_work_area }.uniq

davidgisbey:
Looks good James. Just a couple of comments 👍

paulrobertlloyd (approved):

davidgisbey (approved):
💯 lgtm

PR #640, opened 2021-02-04
Update financial support partial to use standardised content

JR-G:
### Context

This is an additional section on the course details page that is being updated to use standardised content, as part of the work to improve guidance for candidates.

The standardised content box has now been added for the 'Financial support' section:

Screenshot

Prototype: https://find-prototype.herokuapp.com/course/1N1/2XT2

### Guidance to review

This is using the new component & styling added in [#638](https://github.com/DFE-Digital/find-teacher-training/pull/638).

### Trello card

https://trello.com/c/ZX2WSDA1/2938-%F0%9F%97%BA-dev-update-standardised-content-for-financial-support-on-course-detail-pages

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/_fees_and_financial_support.html.erb
+++ b/app/views/courses/_fees_and_financial_support.html.erb
@@ -1,20 +1,22 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-financial-support">Fees and financial support</h2>
 
-  <% if course.has_fees? %>
-    <%= render partial: 'courses/fees' %>
+  <%= render AdviceComponent.new(title: "Financial support") do %>
+      <% if course.salaried? %>
+      <%= render partial: 'courses/financial_support/salaried' %>
+    <% elsif course.excluded_from_bursary? %>
+      <%= render partial: 'courses/financial_support/loan', locals: { course: course } %>
+    <% elsif course.bursary_only? %>
+      <%= render partial: 'courses/financial_support/bursary', locals: { course: course } %>
+    <% elsif course.has_scholarship_and_bursary? %>
+      <%= render partial: 'courses/financial_support/scholarship_and_bursary', locals: { course: course } %>
+    <% else %>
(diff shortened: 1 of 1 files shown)


app/views/courses/financial_support/_advice.html.erb
```diff
@@ -0,0 +1,2 @@
+<p class="govuk-body">You may be eligible for a <a href="https://getintoteaching.education.gov.uk/funding-my-teacher-training/tuition-fee-and-maintenance-loans">loan while you study</a> – note that you’ll have to apply for <a href="https://www.gov.uk/student-finance">undergraduate student finance</a>.</p>
+<p class="govuk-body">Find out about financial support if you’re from <a href="https://getintoteaching.education.gov.uk/explore-my-options/overseas-graduates">outside the UK</a>.</p>
```
paulrobertlloyd, line 2: Lets use the `govuk_link` helper for these links – means they get the right styling.

paulrobertlloyd: Actually, scratch that. We shouldn’t need any new partials as we should be using those we already have in `app/views/courses/financial_support` folder, partials which are shown depending on the course. Where ever the logic for that is, I suspect that’s where the new `AdviceComponent` should be used.

app/components/advice_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<aside class="app-advice">
```
george-bit, line 1: I can see that you have set the base branch correctly on Github, but have you rebased from `2937-add-content-for-quals-to-course-detail-pages`?

The component files files have been added there, and so shouldn't show on this PR. Nothing urgent, but could possibly cause a merge conflict down the road. 😄

george-bit (approved):
Approved pending changes... although obviously we aren't merging just yet!

app/components/advice_component.html.erb
george-bit: Oop i think you have rebased but maybe before I pushed up the last commit where I added the component! 😓

PR #643, opened 2021-02-05
Remove location map from course details page

malcolmbaig:
### Context
We're reworking the locations UX. Part of this design change involves
removing the locations section from the course details page.

- Wrap the `_apply` partial with a new feature flag check.
- Conditionally render the apply button at the bottom of the page if the feature flag is active.
- Conditionally return early from the JS function that initialises the map.
---
#### Feature flag off:

Screenshot_20210208_154419
---
#### Feature flag on:
Screenshot_20210208_154738

### Guidance to review
- Review diff.
### Trello card
https://trello.com/c/h3R0gdap
### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -21,3 +21,4 @@ feature_flags:
   cycle_ending_soon: false
   cycle_has_ended: false
   display_apply_button: true
+  ucas_only_locations: false
--- a/app/views/courses/_apply.html.erb
+++ b/app/views/courses/_apply.html.erb
@@ -8,16 +8,16 @@
   <% if @course.has_vacancies? %>
     <% if FeatureFlag.active?(:display_apply_button) %>
       <p class="govuk-body">
-        <%= render GovukComponent::StartNowButton.new(
-          text: 'Apply for this course',
-          href: apply_path(provider_code: course.provider.provider_code, course_code: course.course_code),
-          html_attributes: { 'data-qa': 'course__apply_link' },
-        ) %>
+      <%= render GovukComponent::StartNowButton.new(
+        text: 'Apply for this course',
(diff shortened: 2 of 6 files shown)


config/settings.yml
```diff
@@ -21,3 +21,4 @@ feature_flags:
   cycle_ending_soon: false
   cycle_has_ended: false
   display_apply_button: true
+  improved_locations: false
```
malcolmbaig, line 24: Is this name adequate? Should it be called something else?

paulrobertlloyd: Perhaps `ucas_only_locations`?

paulrobertlloyd: And maybe that flag should encompass what’s covered in these cards?

* https://trello.com/c/NxrO8GoV
* https://trello.com/c/ErVvUSIA

So, with flag set to `false`:
* as is now (map and codes shown on course detail page).

When set to `true`:
* no map or location codes table shown on course detail page
* selecting UCAS apply route takes you to UCAS apply page (https://trello.com/c/ErVvUSIA)
* the UCAS apply page shows the location codes table (https://trello.com/c/NxrO8GoV)

Essentially, these 3 changes constitute one feature.

malcolmbaig: I've noticed that the card for this PR doesn't mention hiding the location codes table or changing the position of the apply button (as per the design history entry). I don't see that carded up elsewhere, and it makes sense to include those changes here.

Will reopen this PR once those are accounted for. cc: @davidgisbey

davidgisbey: I think we really need to wait for the ttw changes before we can remove the locations from the page too.

davidgisbey (approved):
lgtm

PR #644, opened 2021-02-05
Add localised strings for GIT contact numbers and URLs

paulrobertlloyd:
### Context

Currently the contact information for GIT that we show in the footer differs from the information we show on course pages.

Like we’re doing on Apply (https://github.com/DFE-Digital/apply-for-teacher-training/pull/3995), store common GIT values as localised strings. 

### Guidance to review

* ~~To render the `nbsp;` in the `opening_times` string, am using `html_safe` method on `t`; is that okay?~~
* Unlike on Apply, have not moved all GIT URLs into the localisation file – should we?
* Noticed that we’re not using `govuk_link_to`; can address in a later PR

### Trello card

https://trello.com/c/wpgn9P98/

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -69,3 +69,12 @@ en:
   page_titles:
     error_prefix: "Error: "
     subjects_filter: "Select the subjects you want to teach"
+  service_name:
+    get_into_teaching: Get Into Teaching
+  get_into_teaching:
+    # TODO: Update to 0800 389 2500 when GIT leaves beta
+    tel: 0800 389 2501
+    opening_times: "Monday to Friday, 8.30am to 5pm (except public\u00A0holidays)"
+    # TODO: Update domain names when GIT leaves beta
+    url_get_an_advisor: https://beta-adviser-getintoteaching.education.gov.uk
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -69,3 +69,12 @@ en:
   page_titles:
     error_prefix: "Error: "
     subjects_filter: "Select the subjects you want to teach"
+  service_name:
+    get_into_teaching: Get into teaching
+  get_into_teaching:
+    # TODO: Update to 0800 389 2500 when GIT leaves beta
+    tel: 0800 389 2501
+    opening_times: Monday to Friday, 8.30am to 5pm (except public&nbsp;holidays)
```
malcolmbaig, line 77: Unicode might be a better fit here. Means you don't need the HTML entity and can remove the calls to `.html_safe` in the template.

```suggestion
    opening_times: "Monday to Friday, 8.30am to 5pm (except public\u00A0holidays)"
```

paulrobertlloyd: Awesome! This is much better, thanks @malcolmbaig 😄

malcolmbaig (approved):

PR #655, opened 2021-02-11
Fix Apply section bookmark link

TobyRet:
### Context
Clicking on the 'Apply' link in the contents section was not taking the user to the Apply section (when the `ucas_only_locations` feature flag was turned on)

### Changes proposed in this pull request
- Fixed

### Trello card
https://trello.com/c/oplOt4W5/3005-apply-on-course-details-page-does-not-link-to-new-apply-section

### Checklist

- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #656, opened 2021-02-11
Refactor apply button rendering based on ucas_only_locations feature

malcolmbaig:
### Context

https://github.com/DFE-Digital/find-teacher-training/pull/655
Refactors:
- Use _apply_button in the _apply partial as well as the root
  show.html.erb template
- Make the if branching in show.html.erb clearer

Bug fixes:
- Add a container div with the correct id to fix the 'Apply' link at the
  top of the page when ucas_only_locations is switched on
- When the feature is on, show a warning instead of an apply button if
  the course has no vacancies
### Guidance to review
- View diff
### Trello card
https://trello.com/c/oplOt4W5/3005-apply-on-course-details-page-does-not-link-to-new-apply-section
### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/_apply.html.erb
+++ b/app/views/courses/_apply.html.erb
@@ -4,21 +4,7 @@
 
 <div class="govuk-!-margin-bottom-8" id="section-apply">
   <% if @course.has_vacancies? %>
-    <% if FeatureFlag.active?(:display_apply_button) %>
-      <p class="govuk-body">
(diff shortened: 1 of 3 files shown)


TobyRet (approved):

davidgisbey (approved):
Good stuff 👌

PR #659, opened 2021-02-15
Show links to relevant teacher training websites

raamSoftwire:
### Context

We should show different text to users in the devolved nations.

- find the country from the geocode results
- pass it through as a param
- conditionally render different text for each of the devolved nations
- hide search suggestions for devolved nations

### Guidance to review

- Is it OK to pass the country through as a param?
- I intended to leave all other behaviour unaffected, ie suggested search for England or if you search for Rome etc.

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/form_objects/result_filters/location_filter_form.rb
+++ b/app/form_objects/result_filters/location_filter_form.rb
@@ -48,6 +48,7 @@ module ResultFilters
           lng: results.longitude,
           loc: results.address,
           lq: location_query,
+          c: country(results),
         }
       end
     end
@@ -62,6 +63,13 @@ module ResultFilters
 
     def search_radius
       @params[:rad]
+    end
+
+    def country(results)
+      flattened_results = results.address_components.map(&:values).flatten
+      countries = [DEVOLVED_NATIONS, 'England'].flatten
+
+      countries.each { |country| return country if flattened_results.include?(country) }
     end
   end
 end
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -266,6 +266,14 @@ class ResultsView
     end
   end
 
+  def country
+    query_parameters['c']
+  end
+
+  def devolved_nation?
+    DEVOLVED_NATIONS.include?(country)
+  end
+
 private
 
   def nearest_location(course)
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -210,7 +210,7 @@ class ResultsView
   end
 
   def suggested_search_visible?
-    course_count < SUGGESTED_SEARCH_THRESHOLD && suggested_search_links.any?
+    course_count < SUGGESTED_SEARCH_THRESHOLD && suggested_search_links.any? && !devolved_nation?
   end
 
   def suggested_search_links
--- a/app/views/results/index.html.erb
+++ b/app/views/results/index.html.erb
@@ -33,8 +33,7 @@
   <div class="govuk-grid-column-two-thirds">
     <% if @results_view.no_results_found? %>
       <div class="app-search-results">
-        <h2 class="govuk-heading-m">There are no courses matching your&nbsp;search</h2>
-        <%= render partial: 'try_another_search_text' %>
+        <%= render partial: 'no_results', locals: { country: @results_view.country, devolved_nation: @results_view.devolved_nation? } %>
       </div>
     <% else %>
       <% unless @results_view.provider_filter? %>
--- a/spec/view_objects/results_view_spec.rb
+++ b/spec/view_objects/results_view_spec.rb
@@ -444,38 +444,53 @@ describe ResultsView do
   end
 
   describe '#suggested_search_visible?' do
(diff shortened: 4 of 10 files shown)


app/views/results/_no_results.erb
```diff
@@ -0,0 +1,21 @@
+<% if devolved_nation%>
```
davidgisbey, line 1: `<% if devolved_nation %>`

app/views/results/index.html.erb
```diff
@@ -32,8 +32,7 @@
 
   <div class="govuk-grid-column-two-thirds">
     <% if @results_view.no_results_found? %>
-      <h2 class="govuk-heading-l">There are no courses matching your&nbsp;search</h2>
-      <%= render partial: "try_another_search_text" %>
+      <%= render partial: "no_results", locals: {country: @results_view.country, devolved_nation: @results_view.devolved_nation?} %>
```
davidgisbey, line 35: <%= render partial: "no_results", locals: { country: @results_view.country, devolved_nation: @results_view.devolved_nation? } %>

app/form_objects/result_filters/location_filter_form.rb
```diff
@@ -63,5 +64,13 @@ def location_query
     def search_radius
       @params[:rad]
     end
+
+    def country(results)
```
davidgisbey, line 68: I'm wondering there's a decent place create a `DEVOLVED_NATIONS` constant so we can reuse it here
```
flattened_results = results.address_components.map(&:values).flatten
NameofModel::DEVOLVED_NATIONS.each { |nation| return nation if flattened_results.incldue(nation)
```

app/view_objects/results_view.rb
```diff
@@ -258,6 +258,14 @@ def placement_schools_summary(course)
     end
   end
 
+  def country
+    query_parameters['c']
+  end
+
+  def devolved_nation?
```
davidgisbey, line 265: the `DEVOLVED_NATIONS` constant could then re reused here

NameofModel::DEVOLVED_NATIONS.include(country)

davidgisbey (commented):
Looks good overall. A couple of nits and a potential change. Let me know what you think.

app/form_objects/result_filters/location_filter_form.rb
raamSoftwire: Good idea, I've put it in the initialisers, similar to GCSE grades on Apply.

app/view_objects/results_view.rb
raamSoftwire: Done

app/view_objects/results_view.rb
```diff
@@ -210,7 +210,7 @@ def all_subjects
   end
 
   def suggested_search_visible?
-    course_count < SUGGESTED_SEARCH_THRESHOLD && suggested_search_links.any?
+    course_count < SUGGESTED_SEARCH_THRESHOLD && suggested_search_links.any? && !devolved_nation?
```
raamSoftwire, line 213: I still need to write a test for this, but I'm not sure where it should live.

spec/view_objects/results_view_spec.rb
```diff
@@ -795,4 +795,38 @@ def stub_request_with_meta_count(count)
...
+    context 'where country is England' do
+      let(:results_view) { described_class.new(query_parameters: { 'c' => 'England' }) }
+
+      it 'returns false' do
+        expect(results_view.devolved_nation?).to be false
+      end
+    end
+
+    context 'where country is nil' do
+      let(:results_view) { described_class.new(query_parameters: { 'c' => 'nil' }) }
```
raamSoftwire, line 825: I'm not sure if this test is necessary?

spec/views/results/no_results_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+                                href: 'https://www.education-ni.gov.uk/articles/initial-teacher-education-courses-northern-ireland')
+    end
+  end
+
+  context 'England' do
+    let(:html) { render partial: 'results/no_results', locals: { country: 'England', devolved_nation: false } }
+    let(:results_view) { described_class.new(query_parameters: { 'funding' => '8' }) }
+
+    it 'renders try another search text' do
+      expect(html).to match('There are no courses matching your search')
```
raamSoftwire, line 44: This test currently fails because `@results_view` is `nil`. I thought the `let` on line 41 would take care of this, but I must be missing something.

app/view_objects/results_view.rb
raamSoftwire: I've added this in now

spec/views/results/no_results_spec.rb
raamSoftwire: I've now sorted this too.

spec/view_objects/results_view_spec.rb
davidgisbey: Yea i think it's fine in 👍

davidgisbey (approved):
This is really great work 👌 

At first, I thought that passing England through seemed a bit unnecessary as the behaviour is the same way when passing nil (or an address from outside the UK) does, but the more I think about it, I think it's good that it's explicit.

PR #665, opened 2021-02-17
Use notification banner component for flash success messages

paulrobertlloyd:
### Context

We have a success flash message which uses bespoke styling. However, there is now a component in the design system that covers this scenario, and it’s available via the `govuk-components` gem.

* Replace custom alert message markup with `GovukComponent::NotificationBanner`
* Remove `success-summary` styles
* Tweak whitespace in application layout template

| Before | After |
| - | - |
|  |  |

### Guidance to review

* This PR is dependent on https://github.com/DFE-Digital/govuk-components/issues/109 - we need to set `role="alert"`, and the gem doesn’t currently support doing this.

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


stevehook (approved):
Looks good to me when running locally. I guess you still have to upgrade the `govuk-components` gem but what is here so far is all good to me.

PR #669, opened 2021-02-18
Lint ERB files

paulrobertlloyd:
### Context

Currently ERB files don’t need to follow the same rules used elsewhere in the application. Different styles are therefore used throughout the codebase possibly leading to confusion and certainly leading to untidiness!

Follows on from the same work on Apply, i.e. https://github.com/DFE-Digital/apply-for-teacher-training/pull/4060

* Add ERB lint
* Add `lint_erb` task, and add it to default rake task
* Fix all 513 violations 😲

### Guidance to review

* ~~One outstanding violation that I’m not sure how to fix:~~ Fixed. Thanks @stevehook!
 
  ```
  Style/SymbolProc: Pass `&:subject_name` as an argument to `sort_by` instead of a block.
  In file: app/views/result_filters/subject/_form.html.erb:33
  ```

* This does not lint ERB files using custom cops… although if https://github.com/DFE-Digital/apply-for-teacher-training/pull/4095 is merged on Apply, given that we inherit `https://raw.githubusercontent.com/DFE-Digital/apply-for-teacher-training/master/.rubocop.yml`, maybe it will? 🤔 

  I’m unsure what impact that will have on Find… we want to lint for the correct use of `govuk_link_to`, but we currently don’t use `govuk_button_to` or `govuk_submit` on Find, so would want to exclude those checks. Curious to hear what the solution to this might be; a number of override config values in this projects’s `.rubocop.yml`?

### Trello card

https://trello.com/c/Za2awSWs/

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


frankieroberto (approved):
Are these picked up by CodeClimate?

TobyRet:
> CodeClimate?

@frankieroberto - CodeClimate is currently hooked up but TBH its a bit of a headache when it comes to permissions and fixing / won't fix. Publish / Register have moved to https://github.com/apps/sonarcloud  and I was planning to see how if this may be an option for Find moving forwards

frankieroberto:
@TobyRet I helped to set up GitHub actions for code listing in a previous project. Seemed to work well (and free).

TobyRet (approved):
Nice job! Thanks for doing this @paulrobertlloyd

PR #671, opened 2021-02-19
Default filters to show courses with vacancies by default | Fix missing params

TobyRet:
### Context

This PR addresses a couple of issues as they both related to query. 

The first is ensuring that we only show courses with vacancies by default.

The second is a bug fix that occurs when a candidate searches by location, but subsequently courses in the search results are not sorted by distance. This is because the new filters form was not adding all the location params as hidden fields.

- Change the default behaviour of the search functionality to show courses with vacancies only. 
- Add missing parameters as missing fields and add test missing coverage.

### Guidance to review
- Review by commit
- Spin up the app and run any search. You should see results are automatically filtered by courses with vacancies. Any subsequent filtering by location should result in the expected behaviour - courses should be ordered by distance.

### Trello cards
- https://trello.com/c/4AHWjswy/3062-show-vacancies-by-default-on-find
- https://trello.com/c/nlpAebUf/3070-issue-with-the-find-search-results

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/view_objects/result_filters/new_filters_view.rb
+++ b/app/view_objects/result_filters/new_filters_view.rb
@@ -66,6 +66,20 @@ module ResultFilters
       params[:hasvacancies].nil?
     end
 
+    def location_query_params
+      {
+        c: params[:c],
+        l: params[:l],
+        lq: params[:lq],
+        lat: params[:lat],
+        loc: params[:loc],
+        lng: params[:lng],
+        query: params[:query],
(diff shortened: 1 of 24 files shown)


raamSoftwire (approved):
This looks good, thanks Toby. I've tested it locally and can see that the results are sorted by distance and that only courses with vacancies are shown.

Only two tiny notes from me.

PR #674, opened 2021-02-19
Add custom cops and fix resulting violations

paulrobertlloyd:
### Context

Follows on from #669.

As Find’s Rubocop inherits from Apply’s config, and now that Apply has added 3 new custom cops to check for various `govuk_` helpers, we need to:

* Add the same custom cops to this project
* Fix the resulting violations
* Disable checks for use of `govuk_submit` (this project doesn’t use the form builder gem)
* Exclude checks for use of `govuk_link_to` in the application layout (footer links don’t use the `govuk-link` class)

### Guidance to review

For completeness, and given errors will be raised if they are not used, I have also added the following helpers, even though they are currently not used by this project:

* `govuk_button_link_to`
* `govuk_button_to`

Are we fine with this?

### Trello card

https://trello.com/c/Za2awSWs/

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,11 @@ inherit_from:
 Govuk/GovukSubmit:
   Enabled: false
 
+Govuk/GovukLinkTo:
+  Exclude:
+    # link_to in footer
+    - 'app/views/layouts/application.html.erb'
+
 # The following cops were present before adding Apply cops above
 Rails/HasManyOrHasOneDependent:
  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,10 @@
 inherit_from:
   - https://raw.githubusercontent.com/DFE-Digital/apply-for-teacher-training/master/.rubocop.yml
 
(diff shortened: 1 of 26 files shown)


.rubocop.yml
```diff
@@ -1,6 +1,15 @@
...
   - https://raw.githubusercontent.com/DFE-Digital/apply-for-teacher-training/master/.rubocop.yml
 
+# Disabled because this project doesn’t use GOV.UK Design System Form Builder
+Govuk/GovukSubmit:
+  Enabled: false
+
+Govuk/GovukLinkTo:
+  Exclude:
+    # link_to in footer
+    - 'app/views/layouts/application.html.erb'
```
TobyRet, line 11: Is this cop not inherited from Apply (line 2)?

paulrobertlloyd: Do you mean is `'app/views/layouts/application.html.erb'` already excluded in the Apply config? In which case no, because Apply has separate partials for the footer.

TobyRet: 👌

TobyRet (approved):

PR #680, opened 2021-02-24
Wire up provider autocomplete on 'pick a provider' page

stevehook:
### Context

The provider autocomplete is not wired up on the Visit the _try another provider_ page.

Wire up the provider autocomplete using the same markup as elsewhere in the application.

image

### Guidance to review

- How should we test this? It's not covered by the JS tests and we don't have JS enabled feature specs?
- Note that the fix applied to the other provider autocomplete in https://github.com/DFE-Digital/find-teacher-training/pull/679 will also apply to this one.

### Trello card

https://trello.com/c/04BwX3x4/2004-find-provider-autocomplete-not-wired-up

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/result_filters/provider/new.html.erb
+++ b/app/views/result_filters/provider/new.html.erb
@@ -28,7 +28,12 @@
                 <strong>School, university or other training provider</strong>
                 <span class="govuk-hint govuk-!-margin-bottom-0">Enter the name or provider code</span>
               <% end %>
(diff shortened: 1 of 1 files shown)


TobyRet (approved):
Yeah I wonder if there is now a fairly strong case for adding JS enabled feature specs. Worth discussing at next team tech debt session? I would imagine similar considerations could apply to Apply

PR #610, opened 2021-01-15
Use `govuk-components` gem to replace standard markup

stevehook:
### Context
The `govuk-components` gem provides component wrappers for some of the standard markup shared across services. This PR explores using these components in Find.

### Changes proposed in this pull request

So far:
- [x] Add `govuk-components` gem
- [x] Replace the header with the `Header` component
- [x] Replace the call to action link to Apply with a `StartNowButton` component (see visual change below)
- [x] Replace the phase banner with a component
- [x] Replace a couple of warning panels with a component

Not covered here is replacing the Accordion used for subject selection with Accordion component. We have a requirement to render the current selected subject(s) in an expanded section. It doesn't look as though this is possible with the `govuk-components` gem right now, though I think we can raise a PR to add that feature.

All the changes should be invisible except for this one:

Before:
image

After:
image

### Guidance to review

- Per commit recommended

### Trello card
https://trello.com/c/53RWVPVj/2826-dev-use-govuk-view-components-in-find

### Checklist

- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #696, opened 2021-03-04
Attempt to use Accordion component for subject selection

stevehook:
### Context

We have been using `govuk-components` to simplify (and standardise) some of the markup (see https://github.com/DFE-Digital/find-teacher-training/pull/610). Originally we were unable to use the `Accordion` component because it didn't have an `expanded` option. Now that's been added (see https://github.com/DFE-Digital/govuk-components/pull/115) we can used it.

- [x] Replace markup for subject selection Accordion with component.

### Guidance to review

- There should be no visible changes.
- Had to change the `subject_area_is_selected?` helper - I'm not sure how this worked with the old markup either because it was comparing ids with subject codes.
- Also lost a few `data-qa` attributes (because they are not part of the govuk-components markup). This required a few tweaks to the SitePrism page selectors - these are now using a selector based on an existing CSS class rather than a
`data-qa` attribute.

### Trello card
https://trello.com/c/0fMYPksT/2862-dev-use-govuk-view-components-in-find-to-replace-accordion-markup

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/result_filters/subject/_form.html.erb
+++ b/app/views/result_filters/subject/_form.html.erb
@@ -7,17 +7,14 @@
     <%= form_with(url: subject_path, method: :post, data: { 'ga-event-form' => 'Subject' }) do |f| %>
       <%= render 'shared/hidden_fields', exclude_keys: %w[subjects senCourses], form: f %>
       <h1 class="govuk-heading-l" data-qa="heading"><%= I18n.t('page_titles.subjects_filter') %></h1>
-      <div class="accordion govuk-form-group" data-module="govuk-accordion">
+
+      <%= govuk_accordion(id: 'subject-area-accordion') do |accordion| %>
         <% @subject_areas.each_with_index do |subject_area, counter| %>
-          <div class="govuk-accordion__section<%= if subject_area_is_selected?(subject_area: subject_area) || (counter.zero? && no_subject_selected_error?) then ' govuk-accordion__section--expanded' end %>" data-qa="subject_area">
-            <div class="govuk-accordion__section-header">
-              <h2 class="govuk-accordion__section-heading" data-qa="subject_area__name">
-                <button type="button" data-qa="subject_area__accordion_button" id="accordion-heading-<%= subject_area.typename %>" aria-controls="<%= subject_area.typename.downcase %>-content-<%= counter %>" class="govuk-accordion__section-button" aria-expanded="<%= subject_area_is_selected?(subject_area: subject_area) || (counter.zero? && no_subject_selected_error?) ? 'true' : 'false' %>">
-                  <%= subject_area.name %>
-                </button>
-              </h2>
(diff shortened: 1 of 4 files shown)


app/views/result_filters/subject/_form.html.erb
```diff
@@ -7,17 +7,14 @@
...
-                </button>
-              </h2>
-              <span class="govuk-accordion__icon" aria-hidden="true"></span>
-            </div>
+          <%= accordion.add_section(
+            title: subject_area.name,
+            expanded: subject_area_is_selected?(subject_area: subject_area) || (counter.zero? && no_subject_selected_error?),
+            html_attributes: { 'data-qa': 'subject_area' },
+          ) do %>
             <div id="<%= subject_area.typename.downcase %>-content-<%= counter %>" class="govuk-accordion__section-content" aria-labelledby="accordion-heading-<%= subject_area.typename.downcase %>">
```
paulrobertlloyd, line 18: Is `govuk-accordion__section-content` not part of the component already, or are overriding this somehow? Where does the `add_section` method come from? (I had a look at the component code and couldn’t quite work it out).

spec/site_prism/page_objects/page/result_filters/subject_page.rb
```diff
@@ -13,8 +13,8 @@ class Subject < SitePrism::Section
           end
 
           element :legend, '[data-qa="subject_area__legend"]'
-          element :name, '[data-qa="subject_area__name"]'
-          element :accordion_button, '[data-qa="subject_area__accordion_button"]'
+          element :name, '.govuk-accordion__section-heading'
+          element :accordion_button, '.govuk-accordion__section-button'
```
paulrobertlloyd, line 17: I don’t mind this; I’d be happy of we removed many of the `data-qa` attributes we have on Find as they are usually superfluous.

paulrobertlloyd (dismissed):
~~LGTM, just one question about how we’re adding section content via the component.~~

app/views/result_filters/subject/_form.html.erb
paulrobertlloyd: Actually, inspecting the code, and seeing how it renders, we actually have `govuk-accordion__section-content` nested within `govuk-accordion__section-content` which doubles the space between the section header and content:

| Production | This PR |
| - | - |
| Screenshot 2021-03-07 at 14 37 27 | Screenshot 2021-03-07 at 14 37 13 |

Here’s a screenshot of the offending code:

Screenshot 2021-03-07 at 14 37 44

My side senses were right, `govuk-accordion__section-content` shouldn’t feature in any code outside of the component.

stevehook: Good spot, I think that's fixed now

paulrobertlloyd (approved):

PR #701, opened 2021-03-08
Set Cache-Control header in SitemapsController

TobyRet:
### Context
Currently Find makes a very expensive call to the API whenever the sitemap is generated - see https://github.com/DFE-Digital/find-teacher-training/blob/master/app/controllers/sitemaps_controller.rb#L1

- Previously we were caching the API response used to generate the sitemap, however this can be quite inefficient as we may have up to 4 instances of Find running at one time.
- Instead we now cache the sitemap itself by setting the Cache-Control header, which will be respected by CloudFront

### Trello card
https://trello.com/c/sijB88eS/3129-cache-find-sitemap

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/sitemaps_controller.rb
+++ b/app/controllers/sitemaps_controller.rb
@@ -1,18 +1,12 @@
 class SitemapsController < ApplicationController
   def show
-    @courses = fetch_courses
-  end
+    @courses = Course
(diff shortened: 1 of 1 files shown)


vigneshmsft (approved):
🚀

PR #706, opened 2021-03-10
De-emphasise summary labels on the results page

frankieroberto:
The labels within the results list (Course, Financial support, Locations, Vacancies)  are repeated for every item. With the exception of Vacancies (which displays only Yes or No), they're also pretty self-explanatory if you read the values first.

Because of this, perhaps we can de-emphasis them by dropping the bold and using dark grey text, to give more focus to the content which does change for each course?  This is also what the [Teaching vacancies service](https://teaching-vacancies.service.gov.uk/teaching-jobs-in-London) does (albeit using smaller text).

## Before

find-before

## After

find-after

### DIFF
--- a/app/views/results/index.html.erb
+++ b/app/views/results/index.html.erb
@@ -71,7 +71,7 @@
                 <span class="app-search-result__course-name" data-qa="course__name"><%= course.decorate.display_title %></span>
               <% end %>
             </h2>
-            <dl class="app-description-list">
+            <dl class="app-description-list app-description-list--course-result">
               <dt class="app-description-list__label">Course</dt>
               <dd data-qa="course__description"><%= course.description %></dd>
               <% if @results_view.location_filter? && @results_view.has_sites?(course) %>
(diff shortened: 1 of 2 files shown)


paulrobertlloyd:
Interesting. What prompted this change? I’m not opposed to it… might want to check what other description list this change effects, and if we want the same treatment. I think the course summary (and maybe contact details) on a course details page uses the same class.

frankieroberto:
@paulrobertlloyd I had this idea a while back, I think prompted by spending a lot of time browsing through Find and feeling like all the bold was distracting...

I forgot to check the other places that class was used though – moved the CSS to a modifier for now. Here's how those summary lists look if they have the same treatment:

Screenshot 2021-03-10 at 13 33 13

Screenshot 2021-03-10 at 13 33 24

davidgisbey (approved):

frankieroberto:
I'll close this for now, and we can revisit as part of future changes.

PR #708, opened 2021-03-11
Update which placement partial to render for school placements

JR-G:
### Context

Some school direct providers allow you to pick a location (and they list all of their locations), but our guidance says you can’t choose, which contradicts this.

Until we have a revised solution, remove this box from School Direct providers.

The ‘Where you will train‘ box does not appear on School Direct or PG apprenticeship courses

### Trello card

https://trello.com/c/ZMzl1hFe/3156-find-remove-where-you-will-train-grey-box-from-school-direct-providers

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/views/courses/_about_schools.html.erb
+++ b/app/views/courses/_about_schools.html.erb
@@ -1,13 +1,17 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-schools"><%= course.placements_heading %></h2>
   <div data-qa="course__about_schools">
+  <% if course.program_type == 'higher_education_programme' %>
     <%= render AdviceComponent.new(title: 'Where you will train') do %>
-      <% if course.provider_type == 'university' %>
-        <%= render partial: 'courses/placement/hei', locals: { course: course } %>
-      <% else %>
-        <%= render partial: 'courses/placement/scitt', locals: { course: course } %>
-      <% end %>
+      <%= render partial: 'courses/placement/hei', locals: { course: course } %>
     <% end %>
+  <% elsif course.program_type == 'scitt_programme' %>
+    <%= render AdviceComponent.new(title: 'Where you will train') do %>
+      <%= render partial: 'courses/placement/scitt', locals: { course: course } %>
+    <% end %>
+  <% end %>
+  <% if course.how_school_placements_work.present? %>
     <%= markdown(course.how_school_placements_work) %>
+  <%end%>
   </div>
 </div>
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -54,7 +54,7 @@
         <% if course.about_course.present? %>
           <li><%= govuk_link_to 'About the course', '#section-about' %></li>
         <% end %>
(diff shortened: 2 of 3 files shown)


app/views/courses/show.html.erb
```diff
@@ -80,7 +80,7 @@
         <%= render partial: 'courses/about_course' %>
       <% end %>
 
-      <% if course.how_school_placements_work.present? %>
+      <% if course.how_school_placements_work.present? || course.program_type == 'higher_education_programme' || course.program_type == 'scitt_programme' %>
```
paulrobertlloyd, line 83: Given the partial already includes logic for `course.program_type`, not sure it’s needed again here.

davidgisbey (approved):

app/views/courses/_about_schools.html.erb
```diff
@@ -1,13 +1,17 @@
...
-        <%= render partial: 'courses/placement/scitt', locals: { course: course } %>
-      <% end %>
+      <%= render partial: 'courses/placement/hei', locals: { course: course } %>
     <% end %>
+  <% elsif course.program_type == 'scitt_programme' %>
+    <%= render AdviceComponent.new(title: 'Where you will train') do %>
+      <%= render partial: 'courses/placement/scitt', locals: { course: course } %>
+    <% end %>
+  <% end %>
+  <% if course.how_school_placements_work.present? %>
```
paulrobertlloyd, line 13: Given the course page already checks for the presence of `course.how_school_placements_work`, not sure needed here as well.

paulrobertlloyd (changes requested):

app/views/courses/show.html.erb
davidgisbey: If a course is a HE or Scitt program then we want the section to render the gray box with the relevant content.

However, we then need to guard against  `how_school_placements_work` being nil or it breaks. 

The reason we have to somewhat duplicate the logic called is if we remove
`<% if course.how_school_placements_work.present? %>` altogether, then it just renders the title and nothing else for school direct courses with no value for `course.how_school_placements_work.present?`

davidgisbey: let me try and phrase that better

app/views/courses/_about_schools.html.erb
davidgisbey: there's a situation where we end up wanting to render scitt and HE guidance, but `how_school_placements_work = nil`. In this situation, this method will blow up

app/views/courses/show.html.erb
davidgisbey: That didn't make loads of sense, but essentially sometimes we want to render our content even if `course.how_school_placements_work.present? == nil` (in the case of SCITT or HE courses). 

But we don't want to render the section if none of these conditions are true. Otherwise, we just end up with the title and nothing else.

paulrobertlloyd: Ah, cool. In which case, :shipit:

paulrobertlloyd (approved):

PR #709, opened 2021-03-11
[DO NOT MERGE] Add GovUk Frontend Cookie Banner Component and update route

george-bit:
### Context

In order to add more clarity to the cookie banner we need to add the ability to reject cookies. 

This commit adds a component to the front end and Javascript to capture the Reject All Cookies button. Also route for cookies page has been updated to reflect URL on Apply.

image

### Guidance to review

Test in review app using in private browsing to check desired behaviour

### Trello card

https://trello.com/c/slpIdWMK/3060-%F0%9F%8D%AA-update-banner-and-page-to-follow-gds-cookie-guidance-find

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


app/views/layouts/application.html.erb
```diff
@@ -24,9 +24,20 @@
       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
 
-    <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
+    <% unless current_page?(cookie_preferences_path) %>
+      <%= render GovukComponent::CookieBanner.new(title: 'Cookies on Find postgraduate teacher training', html_attributes: { hidden: true, 'aria-label': 'Cookies on Find postgraduate teacher training', 'data-module': 'govuk-cookie-banner', 'data-qa': 'cookie-banner' }) do |cookie_banner| %>
+        <%= cookie_banner.with(:body) do %>
+          <p class="govuk-body">We use cookies to collect information about how you use this service.</p>
```
paulrobertlloyd, line 30: Update to content of banner to reflect GDS guidance

```suggestion
          <p>We use some essential cookies to make this service work.</p>
          <p>We’d also like to use analytics cookies so we can understand how you use the service and make improvements.</p>
```

app/webpacker/scripts/cookie-banner.js
```diff
@@ -16,6 +16,10 @@ export default class CookieBanner {
       '[data-qa="cookie-banner__accept"]'
     );
 
+    this.rejectButton = this.$module.querySelector(
+      '[data-qa="cookie-banner__reject"]'
```
paulrobertlloyd, line 20: This is previous code, but I don’t think it’s very clear – it suggests the data attribute is for testing, but actually it’s required to make the JS work. We have 2 options:

* Use a different data attribute, i.e. `data-accept-cookie="true"` (this is what GOV.UK Notify does)
* Use a class name, i.e. `js-cookie-banner-accept`

Whatever the choice, where we require hooks in the HTML to communicate with JS, we should not be using `data-qa-*`.

paulrobertlloyd (changes requested):
A few content changes and a question about how we hook JS into HTML.

Also… do we need the JS fallback banner?

app/views/layouts/application.html.erb
george-bit: I made the content change, do you have an idea of what the fallback banner content would look like? Would this also be in a banner component?

paulrobertlloyd: See: https://design-system.service.gov.uk/components/cookie-banner/#if-the-user-s-device-is-not-running-javascript

no-js

I’d imagine we can use the same component, just without including the slot for the actions.

paulrobertlloyd:
Also noticed that the recommendations state that you should alert the user when their requested choice has been actioned: https://design-system.service.gov.uk/components/cookie-banner/#when-the-user-accepts-or-rejects-cookies-using-the-cookie-banner

So rather than hide the message after clicking on one of the buttons, we’d need to update the contents of the banner instead:

accepted
rejected

app/views/layouts/application.html.erb
george-bit: Sweet as, thanks Paul 👍

PR #730, opened 2021-03-31
Split course description row

raamSoftwire:

On the results page, we split the course description row into two rows.

Before

image

After

image

### Guidance to review

I've used the `inflections.rb` file so that when we `humanize` the strings like `pgce_with_qts` it knows that they're acronyms and is able to correctly render this as `PGCE with QTS`.

### Trello card

https://trello.com/c/JjxzkEW0/3246-dev-split-course-row-into-qualification-and-study-type-rows

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
+
 # Be sure to restart your server when you modify this file.
 
 # Add new inflection rules using the following format. Inflections
 # are locale specific, and you may define rules for as many different
 # locales as you wish. All of these examples are active by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.plural /^(ox)$/i, '\1en'
-#   inflect.singular /^(ox)en/i, '\1'
-#   inflect.irregular 'person', 'people'
-#   inflect.uncountable %w( fish sheep )
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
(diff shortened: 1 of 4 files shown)


config/initializers/inflections.rb
```diff
@@ -4,12 +4,11 @@
 # Add new inflection rules using the following format. Inflections
 # are locale specific, and you may define rules for as many different
 # locales as you wish. All of these examples are active by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.plural /^(ox)$/i, '\1en'
-#   inflect.singular /^(ox)en/i, '\1'
-#   inflect.irregular 'person', 'people'
-#   inflect.uncountable %w( fish sheep )
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
```
raamSoftwire, line 7: This is the standard way to handle acronyms in Rails. But it is a global change that will affect the entire service, could this cause any issues?

From https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#schema-coursefilter

```
Possible values:

qts
pgce_with_qts
pgde
pgce
pgde_with_qts
```

davidgisbey (approved):
lgtm, might be worth checking that Toby has no issue with the inflections though. Although, i can't think of anywhere it would be an issue.

TobyRet (approved):
Looks good to me @raamSoftwire

PR #756, opened 2021-05-05
Remove where you will train guidance from Educate Teacher Training

davidgisbey:
### Context

This provider does not actually find placements for candidates so the guidance is misleading. They've asked us to remove it.

https://www.find-postgraduate-teacher-training.service.gov.uk/results?l=3&query=Educate+Teacher+Training

- Don't show where you will train default scitt content for Educate Teacher Training

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/_about_schools.html.erb
+++ b/app/views/courses/_about_schools.html.erb
@@ -1,11 +1,11 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-schools"><%= course.placements_heading %></h2>
   <div data-qa="course__about_schools">
-    <% if course.program_type == 'higher_education_programme'  && course.provider.provider_code != 'B31'  %>
+    <% if course.program_type == 'higher_education_programme' && course.provider.provider_code != 'B31'  %>
       <%= render AdviceComponent.new(title: 'Where you will train') do %>
         <%= render partial: 'courses/placement/hei', locals: { course: course } %>
       <% end %>
-    <% elsif course.program_type == 'scitt_programme' %>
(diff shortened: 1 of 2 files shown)


app/views/courses/_about_schools.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-schools"><%= course.placements_heading %></h2>
   <div data-qa="course__about_schools">
-    <% if course.program_type == 'higher_education_programme'  && course.provider.provider_code != 'B31'  %>
+    <% if course.program_type == 'higher_education_programme' && course.provider.provider_code != 'B31'  %>
       <%= render AdviceComponent.new(title: 'Where you will train') do %>
         <%= render partial: 'courses/placement/hei', locals: { course: course } %>
       <% end %>
-    <% elsif course.program_type == 'scitt_programme' %>
+    <% elsif course.program_type == 'scitt_programme'  && course.provider.provider_code != 'E65' %>
```
TobyRet, line 8: Could this logic be moved into the component and unit tested?

i.e. add a `render` method which only renders if conditions are met. Could be easier to extend this way and also reduces the size of `show_spec.rb`

TobyRet: On second read perhaps not :-(

TobyRet (approved):

app/views/courses/_about_schools.html.erb
davidgisbey: Yea this isn't a great solution tbh. Hopefully, this stuff will be ironed out with the work being done with more structured data on publush

PR #759, opened 2021-05-06
Revert to using addListener

TobyRet:
### Context

_This PR addresses a filter issue on IPhone SE - see https://github.com/DFE-Digital/find-teacher-training-prototype/pull/26_

`addListener` has been deprecated in favour of `addEventListener`, as used in the MOJ frontend version of this component: https://github.com/ministryofjustice/moj-frontend/blob/master/src/moj/components/filter-toggle-button/filter-toggle-button.js#L14

This is because `addEventListener` is only supported in Safari for `MediaQueryList` objects from version 14 onwards: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener – and old iPhones are unable to upgrade to this.

`addListener` seems to have been maintained by browsers for backwards-compatibility, at least for now.

A better longer-term fix might be to feature-detect support for `addEventListener` and then fall back to `addListener` if that is not available. Or alternatively to rewrite the javascript so that the component at least "fails safe" and uses the no-javascript mode if any of the component javascript is unsupported.

Use `addListener` instead of `addEventListener` to enable the filters to work with older browsers

### Trello card
https://trello.com/c/mZCYoh0K/3327-bug-find-filters-not-working-correctly-on-iphone-se

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/webpacker/scripts/filter-toggle-button.js
+++ b/app/webpacker/scripts/filter-toggle-button.js
@@ -7,7 +7,7 @@ export const FilterToggleButton = class {
 
   setupResponsiveChecks () {
     this.mq = window.matchMedia(this.options.bigModeMediaQuery)
-    this.mq.addEventListener('change', $.proxy(this, 'checkMode'))
+    this.mq.addListener($.proxy(this, 'checkMode'))
     this.checkMode(this.mq)
   }
 


paulrobertlloyd:
Should we feature test for `addEventListener` and use that if supported, else use `addListener`. Seems like that’d be more robust, and prevent this breaking again in the future. Not sure how we test for support though – @frankieroberto?

frankieroberto:
@paulrobertlloyd I think we should add a tech debt card to review the javascript for this component altogether? We could use feature detection, or a polyfill, or perhaps even move the media query to CSS (although perhaps there's a reason why it wasn’t originally done this way). Currently, if any of the javascript for the component fails, then the filters get stuck in a state with no Close button – whereas ideally it’d fail safe to the no-js version.

Would also be good to remove the dependency on jQuery and move the javascript closer to the standards and conventions of `govuk-frontend`.

For now, I don’t _think_ there are yet any browsers that have dropped support for `addListener` (and perhaps they never will?), and this seems to be what Manage are doing too: https://github.com/DFE-Digital/manage-teacher-training-applications-prototype/blob/master/app/views/_components/filter-toggle-button/filter-toggle-button.js#L14

PR #772, opened 2021-05-12
Only use sites with vacancies for info on results page

davidgisbey:
### Context

At the moment, we only filter by whether a sites site status is new or running before calculating things like the nearest site and how many sites are available.

We should also be filtering out sites with no vacancies.

- Only use sites with vacancies for site related info on the results page 

Locations for course
image

|Before|After|
|------------|-----------|
|image |image|

### Guidance to review

I checked with Paul and he agrees with the general approach. Am i missing anything?

### Trello card

https://trello.com/c/1tPhJM0Z/3391-bug-find-locations-returned-incorrectly-searching-by-town-name-but-not-by-postcode

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/view_objects/results_view.rb
+++ b/app/view_objects/results_view.rb
@@ -162,7 +162,7 @@ class ResultsView
   end
 
   def site_distance(course)
-    distances = new_or_running_sites_for(course).map do |site|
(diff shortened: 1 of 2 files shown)


TobyRet (approved):
Makes sense to me! Nice spot

PR #777, opened 2021-05-17
Display age range on the course details page

JR-G:
### Context

Some providers offer the same course but for different age groups. They input this information on Publish but it is not surfaced on Find automatically. This means that the courses look identical on Find to candidates.

Added the age range metadata to the summary list, on the course details page:

|Before|After|
|---|---|
|image|image|

### Trello card

https://trello.com/c/7GGhiNZp/3387-find-surface-age-range-metadata-from-publish

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -33,6 +33,10 @@
     <dd data-qa="course__qualifications">
       <%= render partial: 'courses/qualification' %>
     </dd>
+    <% if course.age_range_in_years.present? %>
+      <dt class="app-description-list__label">Age range</dt>
+      <dd data-qa="course__age_range"><%= course.age_range_in_years.humanize %></dd>
+    <% end %>
     <% if course.length.present? %>
       <dt class="app-description-list__label">Course length</dt>
       <dd data-qa="course__length"><%= course.length %></dd>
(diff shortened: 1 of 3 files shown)


app/views/courses/show.html.erb
```diff
@@ -33,6 +33,10 @@
     <dd data-qa="course__qualifications">
       <%= render partial: 'courses/qualification' %>
     </dd>
+    <% if course.age_range_in_years.present? %>
+      <dt class="app-description-list__label">Age range</dt>
```
frankieroberto, line 37: Would this be more explicit?

```suggestion
      <dt class="app-description-list__label">Age range taught</dt>
```

davidgisbey (approved):
good to go once Paul has agreed on what age range should be called

defong:
Is this going to back port to publish?

davidgisbey:
What do you mean Sau? It's just data input by a provider on Publish right?

paulrobertlloyd:
@frankieroberto @JR-G I’d be tempted to say *Age range*, if only to avoid being overly specific (training may involve teaching ages above and below this range).

PR #787, opened 2021-05-26
Show a Sandbox phase banner on Sandbox

duncanjbrown:

On Sandbox, show a Sandbox phase banner in the header

Screenshot 2021-05-25 at 21 36 30

### Checklist

- [X] Rebased master
- [X] Cleaned commit history
- [X] Tested by running locally

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -97,8 +97,14 @@
     ) %>
 
     <div class="govuk-width-container">
-      <%= govuk_phase_banner(phase_tag: { text: 'Beta' }) do %>
-        This is a new service – <%= govuk_link_to 'give feedback or report a problem', 'mailto:becomingateacher@digital.education.gov.uk?subject=Feedback%20about%20Find%20postgraduate%20teacher%20training', class: 'govuk-link--no-visited-state' %>
+      <% if Rails.env.sandbox? %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -97,8 +97,14 @@
     ) %>
 
     <div class="govuk-width-container">
-      <%= govuk_phase_banner(phase_tag: { text: 'Beta' }) do %>
-        This is a new service – <%= govuk_link_to 'give feedback or report a problem', 'mailto:becomingateacher@digital.education.gov.uk?subject=Feedback%20about%20Find%20postgraduate%20teacher%20training', class: 'govuk-link--no-visited-state' %>
+      <% if Rails.env.sandbox? %>
+        <%= govuk_phase_banner(phase_tag: { text: 'Sandbox', colour: 'purple' }) do %>
```
vigneshmsft, line 101: could we introduce a new setting? `Settings.banner_tag` for each env?
but doesn't matter if sandbox is the only differentiation we need.

vigneshmsft (approved):

app/views/layouts/application.html.erb
duncanjbrown: It's the colour as well in this case, and only one env, so I think for the time being this is probably the simplest thing that will work.

I do miss `HostingEnvironment` from Apply.

PR #807, opened 2021-06-16
[3512] Send page request events to BigQuery

misaka:
### Context

We want to send a log of all web requests to BigQuery for performance analysis.

Add BigQuery integration, send off all web requests to BigQuery.

### Guidance to review

Code was largely lifted from Apply.

Find does not have a SideKiq instance available to process jobs. We've decided that for the time being, we can probably manage with the built-in async queue handler which processes ActiveJobs concurrently, even if it means losing a small number of requests. This is still a hypothesis, we can add SideKiq processing later as need be.

### Trello card

https://trello.com/c/Fx0UPepe/3512-%F0%9F%8F%88-find-send-page-request-events-to-bigquery

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


misaka:
Failing on rubocop errors. Maybe we need to remove `rubocop-govuk` (see https://github.com/DFE-Digital/apply-for-teacher-training/commit/c0a86461cc3961e1299a07cb8730b2d83d759cb9#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5c)

app/lib/events/web_request.rb
```diff
@@ -13,19 +13,17 @@ def as_json
...
     def with_request_details(rack_request)
-      @event_hash.merge!(
-        request_uuid: rack_request.uuid,
-        request_data: {
-          path: rack_request.path,
-          method: rack_request.method,
-          user_agent: rack_request.user_agent,
-        }
-      )
+      @event_hash[:request_uuid] = rack_request.uuid
```
duncanjbrown, line 16: Out of curiosity, why this change of style? I think I find a hash literal easier to read

spec/support/feature_flag_helper.rb
```diff
@@ -1,9 +1,27 @@
 module FeatureFlagHelper
```
duncanjbrown, line 1: I think we can simplify a bit?

```ruby
module FeatureFlagHelper
  def activate_feature(name)
    initialize_stub
    allow(Settings.feature_flags).to receive(:send).with(name).and_return true
  end

  def deactivate_feature(name)
    initialize_stub
    allow(Settings.feature_flags).to receive(:send).with(name).and_return false
  end

  private

  def initialize_stub
    @_stub_initialized ||= allow(Settings.feature_flags).to receive(:send).and_call_original
  end
end
```

misaka:
> Failing on rubocop errors. Maybe we need to remove `rubocop-govuk` (see [DFE-Digital/apply-for-teacher-training@c0a8646#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5c](https://github.com/DFE-Digital/apply-for-teacher-training/commit/c0a86461cc3961e1299a07cb8730b2d83d759cb9#diff-4f894049af3375c2bd4e608f546f8d4a0eed95464efcdea850993200db9fef5c) and [DFE-Digital/teacher-training-api#1982](https://github.com/DFE-Digital/teacher-training-api/pull/1982))

https://github.com/DFE-Digital/find-teacher-training/pull/809

app/lib/events/web_request.rb
misaka: The `.merge!` were overwriting hash contents and I decided it wasn't worth expending my (very limited) brain power on sorting it out. 😄 Happy to take a suggest code change on this.

spec/support/feature_flag_helper.rb
misaka: If this gets around the issue I mentioned in the [commit message](https://github.com/DFE-Digital/find-teacher-training/pull/807/commits/b884e3e56e0bdca0ce5600db0342036ff1937e41), then sure!

> Needed to change the feature flag helper so that it could handle tests
> that tested code that doesn't check flags other than those that the test
> cares about. Without this change, when non-stubbed features are checked
> for activation, rspec would raise an error.

I'll try it out when I have a moment, or if you have the time feel free too!

duncanjbrown: I tried it out — it does the job of making the suite green!

steventux (changes requested):
Some questions around whether we intend to use the same BQ dataset/table for these events as we do in Apply, I think the way we define the event eg. ` { request_path : "/" }` vs. `{ "request_data" :  { "path" : "/" } }` could be an issue if we are using the same dataset. Does BQ handle this based on `event_type` ?

Also noticed there's inconsistency around the `occurred_at` vs. `timestamp` field in the tests.

app/job/send_event_to_bigquery_job.rb
```diff
@@ -0,0 +1,10 @@
+class SendEventToBigqueryJob < ApplicationJob
+  TABLE_NAME = 'events'.freeze
```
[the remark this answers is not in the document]
misaka: This will be going to a new events table that's specific to Find. This table has a slightly updated format and we'll likely back-port some changes to the Apply events table once things are more settled.

spec/jobs/send_event_to_bigquery_job_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+RSpec.describe SendEventToBigqueryJob do
+  describe '#perform' do
+    let(:request_event) do
+      {
+        environment: 'production',
+        request_method: 'GET',
+        request_path: '/',
+        request_uuid: '1c94ee0c-c217-4c45-a633-d649ff30a4c3',
+        timestamp: Time.zone.now.iso8601,
```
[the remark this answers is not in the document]
misaka: Good spot! Looking into why the test didn't fail there ...

spec/support/feature_flag_helper.rb
misaka: Thanks. Changed.

app/lib/events/web_request.rb
```diff
@@ -0,0 +1,38 @@
+module Events
+  class WebRequest
+    def initialize
+      @event_hash = {
+        occured_at: Time.zone.now.iso8601,
```
steventux, line 5:
```suggestion
        occurred_at: Time.zone.now.iso8601,
```

app/lib/events/web_request.rb
misaka: Figured out the Right :tm: way to do this. Thanks for the prodding.

spec/jobs/send_event_to_bigquery_job_spec.rb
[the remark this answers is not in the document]
misaka: Ah, testing fail. I've fixed this lack by adding a feature test that does a better job of testing the output sent to BigQuery. Thanks!

spec/lib/events/web_request_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      web_request = Events::WebRequest.new
+
+      expect(web_request.as_json['event_type']).to eq 'web_request'
+    end
+
+    it 'sets the occured_at to now' do
+      Timecop.freeze('2021-06-16 12:00:00') do
+        web_request = Events::WebRequest.new
+
+        expect(web_request.as_json['occured_at']).to eq '2021-06-16T12:00:00Z'
```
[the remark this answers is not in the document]
misaka: This is correct, other end of this issue sorted.

app/lib/events/web_request.rb
misaka: 🤦  ... spelling is hard.

spec/support/stub_bigquery_table_requests.rb
```diff
@@ -0,0 +1,11 @@
+module BigQueryTestHelper
```
defong, line 1: casing?

`BigQueryTestHelper` vs `BigqueryTestHelper`

config/initializers/bigquery.rb
```diff
@@ -0,0 +1,12 @@
+require 'google/cloud/bigquery'
+
+BIG_QUERY_API_JSON_KEY = ENV['BIG_QUERY_API_JSON_KEY']
+
+if BIG_QUERY_API_JSON_KEY.present?
```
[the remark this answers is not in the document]
misaka: Ya, I think that's probably a good call.

I'm just wondering if we should rely on the feature flag to be set here too. Or maybe that's the only check we should do here ... if the feature is enabled, then everything has to be set and ready to go.

spec/support/stub_bigquery_table_requests.rb
misaka: Ya, probably.

misaka: Fixed.

config/initializers/bigquery.rb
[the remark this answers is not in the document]
misaka: I've enabled the later open, only initialise when the feature flag is enabled.

app/lib/events/web_request.rb
misaka: Fixed.

steventux (approved):
:+1:

misaka:
@steventux One last change, sorry. I decided to change `s/big_query/bigquery` for consistency, which breaks with the env vars for Apply but is more consistent generally. ¯\_(ツ)_/¯

duncanjbrown (approved):
Thanks @misaka!

PR #817, opened 2021-06-28
Add a heading for Apply section on course pages

paulrobertlloyd:
### Context

Since we decided to always show locations in this section, it is no longer the case that this section shows the button alone. This means we should add a section heading, to differentiate it from the section above.

| Before | After |
| - | - |
| Screenshot 2021-06-28 at 13 22 39 | Screenshot 2021-06-28 at 13 31 37 |
| Screenshot 2021-06-28 at 13 23 10 | Screenshot 2021-06-28 at 13 30 49 |

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


TobyRet (approved):

paulrobertlloyd:
Will merge this now, but will revisit as part of work around showing locations.

PR #818, opened 2021-06-29
[3512] update web_request event format

misaka:

- update the JSON schema for events sent to BigQuery
- add JSON schema validation
- add setting for BigQuery table name

### Guidance to review

Changes tested locally. Also this feature hasn't been enabled in any environments so zero impact to this particular changeset.
 
### Trello card

As the BigQuery event logging hasn't been enabled yet, I've decided to re-use the old Trello ticket number (3512) even though that ticket has been considered "Done".

https://trello.com/c/Fx0UPepe/3512-%F0%9F%8F%88-find-send-page-request-events-to-bigquery

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,9 @@ gem 'jsonapi-deserializable'
 gem 'jsonapi-renderer'
 gem 'jsonapi-serializable'
 
+# Validate JSON schema
+gem 'json-schema'
+
 # Settings for the app
 gem 'config'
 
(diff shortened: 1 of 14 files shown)


steventux (approved):
Only minor question is whether we'd want a separate directory for json schemas but I guess we can do that if and when they become numerous.

misaka:
> Only minor question is whether we'd want a separate directory for json schemas but I guess we can do that if and when they become numerous.

Ya, fair comment. I was just copying what's in Apply, but we should keep it in mind. 👍

PR #832, opened 2021-07-06
enable feature send_web_requests_to_big_query

misaka:
in the envs: qa, staging and production

### Context

These are currently enabled in the secrets, which isn't the right place to do that.

Move the setting for this feature to the repo, which will allow us to remove it from secrets.

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


davidgisbey (approved):
lgtm

PR #833, opened 2021-07-08
[3626] add anonymised user agent and ip to events

misaka:
### Context

We want to have a general sense of the number of unique users in BigQuery events.

Generate `anonymised_user_agent_and_ip` and send to BigQuery.

### Guidance to review

See Trello cards for further info.

### Trello card

https://trello.com/c/hBfnkeMI/3626-add-user-anonymous-identifier-field-to-event-being-sent-to-bq

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/lib/events/web_request.rb
+++ b/app/lib/events/web_request.rb
@@ -47,8 +47,6 @@ module Events
     def anonymised_user_agent_and_ip(rack_request)
       if rack_request.remote_ip.present?
         anonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s)
-      else
-        ''
       end
     end
 
--- a/app/lib/events/web_request.rb
+++ b/app/lib/events/web_request.rb
@@ -20,7 +20,7 @@ module Events
         request_user_agent: rack_request.user_agent,
         request_referer: rack_request.referer,
         request_query: query_to_kv_pairs(rack_request.query_string),
-        anonymised_user_agent_and_ip: anonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s),
+        anonymised_user_agent_and_ip: anonymised_user_agent_and_ip(rack_request),
       )
 
       self
@@ -41,6 +41,14 @@ module Events
       vars = Rack::Utils.parse_query(query_string)
       vars.map do |(key, value)|
         { 'key' => key, 'value' => Array.wrap(value) }
+      end
+    end
+
+    def anonymised_user_agent_and_ip(rack_request)
+      if rack_request.remote_ip.present?
+        anonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s)
+      else
+        ''
       end
     end
 
--- a/app/lib/events/web_request.rb
+++ b/app/lib/events/web_request.rb
@@ -20,6 +20,7 @@ module Events
         request_user_agent: rack_request.user_agent,
         request_referer: rack_request.referer,
         request_query: query_to_kv_pairs(rack_request.query_string),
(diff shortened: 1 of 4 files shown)


app/lib/events/web_request.rb
```diff
@@ -42,5 +43,17 @@ def query_to_kv_pairs(query_string)
         { 'key' => key, 'value' => Array.wrap(value) }
       end
     end
+
+    def anonymised_user_agent_and_ip(rack_request)
+      if rack_request.remote_ip.present?
+        anonymise(rack_request.user_agent.to_s + rack_request.remote_ip.to_s)
+      else
+        ''
```
stevenleggdfe, line 51: @misaka I don't code Ruby, but is '' the same as null?

misaka: Ah, ya, I made the quick assumption that it essentially gets serialised to the same thing when sent over to BigQuery, but thinking about it now I don't think that's the case. Will fix, thanks!

misaka: [`b0fa8ba` (#833)](https://github.com/DFE-Digital/find-teacher-training/pull/833/commits/b0fa8baa8990543eccfa077071f90a57a609f56d)

spec/lib/events/web_request_spec.rb
```diff
@@ -71,6 +83,43 @@
       )
     end
 
+    describe 'anonymised_user_agent_and_ip' do
+      context 'when user agent and ip address both present' do
+        it 'is set to a hash of user agent and ip' do
+          expect(web_request.as_json['anonymised_user_agent_and_ip']).to(
+            eq(Digest::SHA2.hexdigest(user_agent + remote_ip)),
```
duncanjbrown, line 90: as `user_agent` and `remote_ip` are determined, could we put a literal hash in here rather than mirroring the code inside the class?

duncanjbrown (commented):
Looking great — one nit inline, and should we check against/update the schema?

spec/lib/events/web_request_spec.rb
misaka: Sure.

misaka: [`e9159ef` (#833)](https://github.com/DFE-Digital/find-teacher-training/pull/833/commits/e9159efb9b0275aaba7f87dccc8b4b67c58d4cd0)

misaka:
> Looking great — one nit inline, and should we check against/update the schema?

Ya, good point. I'm surprised the schema checker allowed the additional parameter, do you know how to disable that? I'm just checking the JSON schema docs.

misaka:
> > Looking great — one nit inline, and should we check against/update the schema?
> 
> Ya, good point. I'm surprised the schema checker allowed the additional parameter, do you know how to disable that? I'm just checking the JSON schema docs.

Ah, it's [additionalProperties](http://json-schema.org/understanding-json-schema/reference/object.html?highlight=additional#additionalproperties).

spec/lib/events/web_request_spec.rb
misaka: https://github.com/DFE-Digital/find-teacher-training/pull/833/commits/b43ca16f9483149277e8b098773eac8bc8a4df96

duncanjbrown:
> > > Looking great — one nit inline, and should we check against/update the schema?
> 
> > 
> 
> > Ya, good point. I'm surprised the schema checker allowed the additional parameter, do you know how to disable that? I'm just checking the JSON schema docs.
> 
> 
> 
> Ah, it's [additionalProperties](http://json-schema.org/understanding-json-schema/reference/object.html?highlight=additional#additionalproperties).
> 
> 
> 
> 

Nice - I'll make sure the Apply checker does similar!

duncanjbrown (approved):

PR #834, opened 2021-07-09
Add an EOC timetable to Find

davidgisbey:
### Context

Going forward we want to use a CycleTimetable to control the changes in behaviour in discreet phases of the cycle on Find such as the banners, removing the apply button and restricting access to courses when Find is closed.

This PR adds the CycleTimetable.

Going forward this will be used to implement all 4 cards captured in this epic https://trello.com/c/SZRVll8g/2302-%F0%9F%94%81-eoc-dev-find-changes-required-following-design-review

- Add the cycle timetable 

This needs the following values 

1. find opens
2. apply opens (for content on the banner)
3. deadline banner shows
4. apply 1 deadline 
5. apply 2 deadline
6. find_closes

### Guidance to review

Does this all make sense? I think you'd probs eed to have a look at the banner cards for context.

### Trello card

https://trello.com/c/3MhlVfCk/3591-%F0%9F%94%81-find-eoc-dev-phase-1-banners-on-find-to-notify-candidates-the-cycle-will-end

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/cycle_timetable.rb
+++ b/app/services/cycle_timetable.rb
@@ -1,0 +1,100 @@
+class CycleTimetable
+  CYCLE_DATES = {
+    2021 => {
+      find_opens: DateTime.new(2020, 10, 6, 9),
+      apply_opens: DateTime.new(2020, 10, 13, 9),
+      first_deadline_banner: DateTime.new(2021, 7, 12, 9),
+      apply_1_deadline: DateTime.new(2021, 8, 24, 18),
+      apply_2_deadline: DateTime.new(2021, 9, 20, 18),
+      find_closes: DateTime.new(2021, 10, 3),
+    },
+    2022 => {
+      find_opens: DateTime.new(2021, 10, 5, 9),
+      apply_opens: DateTime.new(2021, 10, 12, 9),
+      # the dates from below here are not the finalised but are required for
+      # the current implementation
+      first_deadline_banner: DateTime.new(2022, 7, 12, 9),
+      apply_1_deadline: DateTime.new(2022, 8, 24, 18),
+      apply_2_deadline: DateTime.new(2022, 9, 20, 18),
+      find_closes: DateTime.new(2022, 10, 3),
+    },
+  }.freeze
+
+  def self.current_year
+    now = DateTime.now
+
+    CYCLE_DATES.keys.detect do |year|
+      return year if last_recruitment_cycle_year?(year)
+
+      now.between?(CYCLE_DATES[year][:find_opens], CYCLE_DATES[year + 1][:find_opens])
+    end
+  end
+
+  def self.next_year
+    current_year + 1
+  end
+
+  def self.find_closes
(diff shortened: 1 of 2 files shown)


app/services/cycle_timetable.rb
```diff
@@ -0,0 +1,100 @@
+class CycleTimetable
+  CYCLE_DATES = {
+    2021 => {
+      find_opens: DateTime.new(2020, 10, 6, 9, 0, 0),
+      apply_opens: DateTime.new(2020, 10, 13, 9, 0, 0),
+      deadline_banner: DateTime.new(2021, 7, 12, 9, 0, 0),
```
TobyRet, line 6: Wonder if for consistency we should use event-y type language.
e.g. 'course_search_deadline` - or something. TBH I'm not sure which deadline it refers to ha ha

app/services/cycle_timetable.rb
```diff
@@ -0,0 +1,100 @@
...
+      apply_1_deadline: DateTime.new(2021, 8, 24, 18, 0, 0),
+      apply_2_deadline: DateTime.new(2021, 9, 20, 18, 0, 0),
+      find_closes: DateTime.new(2021, 10, 3, 0, 0, 0),
+    },
+    2022 => {
+      find_opens: DateTime.new(2021, 10, 5, 9, 0, 0),
+      apply_opens: DateTime.new(2021, 10, 12, 9, 0, 0),
+      # the dates from below here are not the finalised but are required for
+      # the current implementation
+      deadline_banner: DateTime.new(2022, 7, 12, 9, 0, 0),
```
TobyRet, line 16: Don't think you need all the additional zeroes in each DateTime? 🤔

TobyRet (approved):
Couple of small comments otherwise LGTM

app/services/cycle_timetable.rb
```diff
@@ -0,0 +1,100 @@
...
+  def self.apply_reopens
+    date(:apply_opens, next_year)
+  end
+
+  def self.preview_mode?
+    Time.zone.now.between?(apply_2_deadline, find_closes)
+  end
+
+  def self.find_down?
+    return false if Time.zone.now > find_opens && Time.zone.now < find_closes
```
JR-G, line 70: This was updated in the Apply class to:
`Time.zone.now.between?(find_closes.end_of_day, find_reopens.beginning_of_day)`

app/services/cycle_timetable.rb
davidgisbey: Yea good point. It's basically just when the deadline banners should start rendering so i didn't have to do something like. apply_1_deadline - 6.weeks or w/e the calculation would be for it starting to show on Monday.

app/services/cycle_timetable.rb
davidgisbey: So i do that it tries to look for it in the 2023 cycle for some reason.

PR #839, opened 2021-07-12
Update the content on the cycle has ended page

davidgisbey:
### Context

As part of the changes we're making to Find for EOC, we need to update the content on the cycle has ended page.

This should be done using the CycleTimetable so it is future proof and does not need to be hardcoded.

Update the content on the cycle has ended page
 
image

### Trello card

https://trello.com/c/mloDU6fr/3593-%F0%9F%94%81-find-eoc-phase-4-find-is-closed-page-including-content-changes-from-last-year

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


EmmaFrith (approved):
content looks good thanks @davidgisbey :)

PR #840, opened 2021-07-12
Apply fix for banner not rendering due to start wizard

davidgisbey:
### Context

When a candidate arrives on find, a start wizard is passed into the flash. This is used to retain state on whether a candidate is moving through the initial flow, or whether they are editing a previous selection. 

We need to fix this, but it will take a while so for now, i've remove the start_wizard from the flash before passing it into the deadline banner component. 

- Render the deadline banner if the only flash is the start_wizard

### Guidance to review

As i've said above, this is a temporary fix. We should rework the way the wizard works. Unfortunately it's not as easy as just passing the params in as they views are completely shared. We can definitely take logic out the partials etc. but it's going to take a bit of time.

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -120,8 +120,8 @@
             <%= notification_banner.slot(:heading, text: flash[:success]) %>
           <% end %>
         <% end %>
-        
-        <%= render DeadlineBannerComponent.new(flash_empty: flash.empty?) unless request.url.include?('/results/filter/subject')%>
+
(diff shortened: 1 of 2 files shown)


frankieroberto (approved):
Seems fine as a quick fix. 👍 

I think I would’ve been tempted to add the call to `<%= render DeadlineBannerComponent.new %>` on the 3 relevant pages (home, results, course) rather than in the layout file.

PR #841, opened 2021-07-12
Default Time to London and update the timetable to use Time.zone.local

davidgisbey:
### Context

At the moment, the CycleTimetable is an hour out of date as the zone has not been set to London.

This PR sets Time.zone to London in the application file and sets the timetable to use Time.zone.local

### Guidance to review

Does this approach seem reasonable?

### Trello card

https://trello.com/c/IxL0GXY0/3634-find-cycletimetable-times-are-an-hour-out-due-to-time-zone

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/cycle_timetable.rb
+++ b/app/services/cycle_timetable.rb
@@ -1,27 +1,27 @@
 class CycleTimetable
   CYCLE_DATES = {
     2021 => {
-      find_opens: DateTime.new(2020, 10, 6, 9),
-      apply_opens: DateTime.new(2020, 10, 13, 9),
-      first_deadline_banner: DateTime.new(2021, 7, 12, 9),
-      apply_1_deadline: DateTime.new(2021, 9, 7, 18),
-      apply_2_deadline: DateTime.new(2021, 9, 20, 18),
-      find_closes: DateTime.new(2021, 10, 3),
+      find_opens: Time.zone.local(2020, 10, 6, 9),
+      apply_opens: Time.zone.local(2020, 10, 13, 9),
+      first_deadline_banner: Time.zone.local(2021, 7, 12, 9),
+      apply_1_deadline: Time.zone.local(2021, 9, 7, 18),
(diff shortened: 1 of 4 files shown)


app/services/cycle_timetable.rb
```diff
@@ -1,27 +1,27 @@
...
+      find_opens: Time.zone.local(2021, 10, 5, 9),
+      apply_opens: Time.zone.local(2021, 10, 12, 9),
       # the dates from below here are not the finalised but are required for
       # the current implementation
-      first_deadline_banner: DateTime.new(2022, 7, 12, 9),
-      apply_1_deadline: DateTime.new(2021, 9, 7, 18),
-      apply_2_deadline: DateTime.new(2022, 9, 20, 18),
-      find_closes: DateTime.new(2022, 10, 3),
+      first_deadline_banner: Time.zone.local(2022, 7, 12, 9),
+      apply_1_deadline: Time.zone.local(2021, 9, 7, 18),
```
TobyRet, line 17: This should be 2022? (`apply_1_deadline:` 2022)

TobyRet (changes requested):
One the dates in the cycle timetable is incorrect otherwise LGTM

spec/features/events/send_web_request_to_big_query_spec.rb
```diff
@@ -33,7 +33,7 @@
           [{
             'environment' => 'test',
             'event_type' => 'web_request',
-            'occurred_at' => Time.now.utc.iso8601,
+            'occurred_at' => Time.zone.now.iso8601,
```
duncanjbrown, line 36: @misaka fyi, changing from UTC! (we already use `Time.zone.local` in Apply)

duncanjbrown (commented):
LGTM pending @TobyRet's comment!

app/services/cycle_timetable.rb
davidgisbey: Good spot!

JR-G (approved):

PR #851, opened 2021-07-22
Add visa sponsorship info to the course page

stevehook:
### Context

We added provider visa sponsorship information to TTAPI/Publish. This PR exposes that information for candidates using Find on the course show page.

- [x] Add a new _International students_ section to the course show page that is only displayed for courses in recruitment cycle 2022 or later.
- [x] Extract logic and markup for this section to a component? (rather than a partial/helper)

### Guidance to review

- Does it make sense to convert the UI for this to a component? Should we be aiming to do that for each section on the Course show page? Or leave as is?

### Trello card

https://trello.com/c/d2xHXK0G/3443-📐🌐-add-international-and-visa-sponsorship-information-to-the-find-course-page

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,10 +5,7 @@ require_relative 'config/application'
 
 Rails.application.load_tasks
 
-# rubocop:disable Rake/Desc
 task lint_ruby: ['lint:ruby']
 task lint_scss: ['lint:scss']
 task lint_erb: ['lint:erb']
-# rubocop:enable Rake/Desc
-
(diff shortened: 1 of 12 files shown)


TobyRet (approved):
LGTM @stevehook - I think using a component here makes sense given it has conditionals.

TobyRet:
> LGTM @stevehook - I think using a component here makes sense given it has conditionals.

Also, given that find currently tests a lot of the view logic via feature specs, trying to move to components / unit testing as we go would be a good move.

PR #854, opened 2021-07-23
Add structured qualification entry requirements to course page

stevehook:
### Context

We added information about degree and GCSE entry requirements to TTAPI/Publish. This PR exposes that information for candidates using Find on the course show page..

- [x] Add a `Course::EntryRequirementsComponent` to replace the existing partial from 2022.
- [x] Render structured GCSE requirements
- [x] Render structured degree requirements
image

- [ ] Change base to master

### Guidance to review

- Does it make sense to extract this to a component and does it make sense to include both the degree and GCSE requirements in the same component (so that it is a drop in replacement for the existing `` partial)?
- This PR duplicates a fair bit of code from Publish (all the conditional rendering logic here is borrowed from the course preview page in Publish). I think we have to live with a certain amount of this while Publish, TTAPI and Find are all separate applications(?).

### Trello cards

https://trello.com/c/Hw7vFgH8/3537-📐-dev-add-structured-gcse-entry-requirements-to-find-course-page
https://trello.com/c/NQPyKmg7/3536-📐-dev-add-structured-degree-entry-requirements-to-find-course-page

### Checklist

- [ ] Rebased master
- [x] Cleaned commit history


app/components/courses/international_students_component.rb
```diff
@@ -0,0 +1,21 @@
...
+    def visa_sponsorship_status
+      if @provider.can_sponsor_all_visas?
+        'can sponsor Student and Skilled Worker visas, but this is not guaranteed'
+      elsif @provider.can_only_sponsor_student_visa?
+        'can sponsor Student visas, but this is not guaranteed'
+      elsif @provider.can_only_sponsor_skilled_worker_visa?
+        'can sponsor Skilled Worker visas, but this is not guaranteed'
+      else
+        'cannot sponsor visas'
+      end
```
frankieroberto, line 18: This content is slightly out-of-date from the [prototype](https://find-prototype.herokuapp.com), which uses this wording (and includes a link):

"We are unable to sponsor a visa. You will need to [get the right visa or status to study in the UK](https://www.gov.uk/government/publications/train-to-teach-in-england-non-uk-applicants/train-to-teach-in-england-non-uk-applicants#visas-and-immigration)."

or

"We can [sponsor Student visas](https://www.gov.uk/government/publications/train-to-teach-in-england-non-uk-applicants/train-to-teach-in-england-non-uk-applicants#visas-and-immigration), but this is not guaranteed."

Also, the logic of which to display need to take the course type (fee-paying or salaried) into account, ie:

```
if course == salaried && provider_can_sponsor_skilled_worker_visa
  // can sponsor Skilled worker visa
elsif course == fee_paying && provider_can_sponsor_student_visa
  // can sponsor Skilled worker visa
else
  // cannot sponsor visa
endif
```

app/components/courses/international_students_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-!-margin-bottom-8">
+  <h2 class="govuk-heading-l" id="section-international-students">International students</h2>
+  <div data-qa="course__international_students">
+    <p class="govuk-body">We <%= visa_sponsorship_status %>.</p>
```
frankieroberto, line 4: Suggestion: move the "we" and the full stop into the conditional text (even if they all currently use this).

app/components/courses/international_students_component.rb
stevehook: Thats updated now. The new logic means that we never show _We can sponsor Student and Skilled Worker visas_. I think it needs an update to publish as well to keep things in sync...

spec/components/courses/entry_requirements_component_spec.rb
```diff
@@ -92,4 +92,23 @@
       )
     end
   end
+
+  context 'when the provider requires a 2:2 and specifies addtional requirements', type: :component do
```
davidgisbey, line 96: 'when the provider requires a 2:2 and specifies additional requirements'

davidgisbey (approved):
gave it a pretty thoroughly test locally and it's working beautifully. Great work!

app/components/courses/entry_requirements_component.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+    <p class="govuk-body">
+      <%= gcse_equivalency_content(course) %>.
+    </p>
+
+    <% if course.additional_gcse_equivalencies %>
+      <p class="govuk-body">
+        <%= course.additional_gcse_equivalencies %>.
+      </p>
+    <% end %>
```
frankieroberto, line 31: When I was testing this locally, I noticed that one course showed a full stop on its own, because the `additional_gcse_equivalencies` field was blank rather than `null`. Think we should either check for presence, or coerce blanks to nulls at the Publish end or in the API?

Also, _hopefully_ providers will include their own full stops... 🤞  – if they don't, or if it's inconsistent, we could consider adding one if not already present (or making the guidance in Publish clearer).

```suggestion
    <% if course.additional_gcse_equivalencies.present? %>
      <p class="govuk-body">
        <%= course.additional_gcse_equivalencies %>
      </p>
    <% end %>
```

app/components/courses/entry_requirements_component.rb
```diff
@@ -0,0 +1,67 @@
...
+        'Candidates with pending GCSEs will not be considered'
+      end
+    end
+
+    def gcse_equivalency_content(course)
+      return 'Equivalency tests will not be accepted' unless course.accept_gcse_equivalency
+
+      case equivalencies.count
+      when 0
+        ''
```
frankieroberto, line 48: In theory we shouldn't need this condition, as the validation will prevent there being no equivalency subjects accepted if the "accept equivalencies" option is ticked. Could remove and allow the default (null?) to be returned?

app/components/courses/entry_requirements_component.rb
```diff
@@ -0,0 +1,67 @@
...
+    def pending_gcse_content(course)
+      if course.accept_pending_gcse
+        'Candidates with pending GCSEs will be considered'
+      else
+        'Candidates with pending GCSEs will not be considered'
+      end
+    end
+
+    def gcse_equivalency_content(course)
+      return 'Equivalency tests will not be accepted' unless course.accept_gcse_equivalency
```
frankieroberto, line 44: Could add a question mark to the end of this method to help indicate that it's a boolean? (Rails magic)

```suggestion
      return 'Equivalency tests will not be accepted' unless course.accept_gcse_equivalency?
```

app/components/courses/entry_requirements_component.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def equivalencies
+      {
+        english: course.accept_english_gcse_equivalency.present?,
+        maths: course.accept_maths_gcse_equivalency.present?,
+        science: course.accept_science_gcse_equivalency.present?,
+
+      }.select { |_k, v| v }.keys
+    end
```
frankieroberto, line 65: I think the capitalisation logic is wrong here, as it is always capitalising the first subject, but not the rest, whereas we only want to capitalise English (if included), as that's a proper noun.

I would suggest doing something like this, which also uses the `to_sentence` Rails helper to avoid the extra condition:

```ruby

def gcse_equivalency_content(course)

  if course.accept_gcse_equivalency?
    "Equivalency tests will be accepted in #{equivalencies}"
  else
    'Equivalency tests will not be accepted'
  end
end

def equivalencies
  subjects = []
  subjects << "English" if course.accept_english_gcse_equivalency.present?
  subjects << "maths" if course.accept_maths_gcse_equivalency.present?
  subjects << "science" if course.accept_science_gcse_equivalency.present?
  
  subjects.to_sentence(last_word_connector: " and ")
end
```

frankieroberto (changes requested):
Looking good. Left a couple of minor things to fix.

frankieroberto (approved):

PR #866, opened 2021-08-04
Add a cycle switcher to Find

JR-G:
### Context

We're no longer using feature flags to move between phases of the cycle on Find to test EoC behaviour & content.

Apply has a switcher which allows you to move between the discreet phases of the cycle – this is to port the behaviour over to Find.

Adds a cycle switcher under the `/cycles` endpoint:
image

This will jump Find forward/back to different points in the cycle. It can only be used in a non-prod environment.

### Guidance to review

Similar to the logic on Apply

Using `http://localhost:3002/cycles` locally you can now switch between the different phases and then view the changes.

### Trello card

https://trello.com/c/K8cXWM7N

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/switcher_controller.rb
+++ b/app/controllers/switcher_controller.rb
@@ -1,4 +1,5 @@
 class SwitcherController < ApplicationController
+  skip_before_action :redirect_to_cycle_has_ended_if_find_is_down
   before_action redirect_to root_path if Rails.env.production?
 
   def cycles; end
@@ -6,8 +7,6 @@ class SwitcherController < ApplicationController
   def update
     new_cycle = params[:change_cycle_form][:cycle_schedule_name]
     SiteSetting.set(name: 'cycle_schedule', value: new_cycle)
-    Rails.application.reload_routes!
-
     flash[:success] = 'Cycle schedule updated'
     redirect_to cycles_path
   end
--- a/app/controllers/switcher_controller.rb
+++ b/app/controllers/switcher_controller.rb
@@ -1,0 +1,14 @@
+class SwitcherController < ApplicationController
+  before_action redirect_to root_path if Rails.env.production?
+
+  def cycles; end
+
+  def update
+    new_cycle = params[:change_cycle_form][:cycle_schedule_name]
+    SiteSetting.set(name: 'cycle_schedule', value: new_cycle)
+    Rails.application.reload_routes!
+
+    flash[:success] = 'Cycle schedule updated'
+    redirect_to cycles_path
+  end
+end
--- a/app/lib/hosting_environment.rb
+++ b/app/lib/hosting_environment.rb
@@ -1,0 +1,5 @@
+module HostingEnvironment
+  def self.production?
+    Rails.env.production?
+  end
+end
--- a/app/models/site_setting.rb
+++ b/app/models/site_setting.rb
@@ -1,0 +1,11 @@
+require './app/lib/redis'
+
+class SiteSetting < Base
+  def self.cycle_schedule
+    RedisService.new.get('cycle_schedule')&.to_sym || :real
+  end
+
+  def self.set(name:, value:)
+    RedisService.new.set(name, value)
+  end
+end
--- a/app/views/switcher/cycles.html.erb
+++ b/app/views/switcher/cycles.html.erb
@@ -1,0 +1,42 @@
+<%= content_for :page_title, 'Recruitment cycles' %>
+
+<% unless HostingEnvironment.production? %>
+  <%= form_with model: ChangeCycleForm.new, url: switch_cycle_schedule_path, method: :post do |f| %>
+    <%= f.govuk_radio_buttons_fieldset :cycle_schedule_name,
+      legend: { text: 'Current point in the recruitment cycle' } do %>
+
+      <%= f.govuk_radio_button :cycle_schedule_name, 'real', label: { text: t('cycles.real.name') }, hint: { text: t('cycles.real.description') } %>
+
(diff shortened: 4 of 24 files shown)


app/controllers/switcher_controller.rb
```diff
@@ -0,0 +1,14 @@
+class SwitcherController < ApplicationController
+  before_action redirect_to root_path if Rails.env.production?
+
+  def cycles; end
+
+  def update
+    new_cycle = params[:change_cycle][:cycle_schedule_name]
+    SiteSetting.set(name: 'cycle_schedule', value: new_cycle)
+    Rails.application.reload_routes!
```
davidgisbey, line 9: So i'm kind of split on this. We definitely could just update the application controller to redirect to the correct page based on the CycleTimetable

i.e. when find is closed always redirect to the find is closed page, otherwise if you hit the find is closed page then redirect to the root path.

Interested to hear what other people think.

davidgisbey:
I think we should probs add a BDD style feature test to this since it's untested atm.

We can just change to each visit the homepage and check the correct banner is rendering/you arrive on the find is closed page on when find is closed

spec/components/summary_list_component_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+describe SummaryListComponent, type: :component do
+  it 'renders component with correct structure' do
+    rows = [
+      key: 'Name:',
+      value: 'Lando Calrissian',
+      action: 'Name',
+      change_path: '/some/url',
+    ]
+    # binding.pry
```
TobyRet, line 11: `# binding.pry`

TobyRet:
> I think we should probs add a BDD style feature test to this since it's untested atm.
> 
> We can just change to each visit the homepage and check the correct banner is rendering/you arrive on the find is closed page on when find is closed

Yeah agree - think we need to config Redis per environment though in order for this to be possible

app/controllers/switcher_controller.rb
TobyRet: I think if the ApplicationController option is simpler and both suits the current use case and is easy to extend (if / when required) then that may be the way to go. Either way might be good to have some docs

app/views/switcher/cycles.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  'Current cycle year' => RecruitmentCycle.current_year,
+  'Next cycle year' => RecruitmentCycle.next_year,
+}) %>
+
+<h2 class="govuk-heading-m">Deadlines</h2>
+
+<p class="govuk-body">(Today is <%= Time.zone.today.to_s(:govuk_date) %>)</p>
+
+<%= render SummaryListComponent.new(rows: {
+  'Appy 1 deadline' => CycleTimetable.apply_1_deadline.to_s(:govuk_date),
```
duncanjbrown, line 37:
```suggestion
  'Apply 1 deadline' => CycleTimetable.apply_1_deadline.to_s(:govuk_date),
```

app/lib/hosting_environment.rb
```diff
@@ -0,0 +1,9 @@
+module HostingEnvironment
+  def self.environment_name
+    ENV.fetch('HOSTING_ENVIRONMENT_NAME', 'unknown-environment')
+  end
+
+  def self.production?
+    environment_name == 'production'
```
duncanjbrown, line 7: this way is preferable and we should move towards it, but as Find uses Rails environments for qa, sandbox etc right now, could this be `Rails.env.production?` and lose the ENV var above?

duncanjbrown (commented):
I agree taking the logic out of `routes.rb` and putting into a filter on the controller is neater. +1 to spec!

app/models/site_setting.rb
```diff
@@ -2,10 +2,18 @@
 
 class SiteSetting < Base
   def self.cycle_schedule
-    RedisService.new.get('cycle_schedule')&.to_sym || :real
+    if HostingEnvironment.known_environment?
+      RedisService.new.get('cycle_schedule')&.to_sym || :real
+    else
+      :real
+    end
```
JR-G, line 9: @davidgisbey , @duncanjbrown this is an experimental commit - the build is failing due to it trying to connect to redis whilst building. This approach gets around that problem (but has issues of its own).

Any suggestions to improving this and/or alternative approaches?

davidgisbey (approved):
We need to pick up the bug with the apply button not showing for Find reopening, but let's do that in a separate PR

config/routes.rb
```diff
@@ -8,22 +8,20 @@
...
 
   get '/cycle-ending-soon', to: redirect('/', status: 301)
   # During the cycle there is no need for a separate path for
   # start#location but this was the root path in the
   # legacy c# app so we're redirecting to root
   get '/start/location', to: redirect('/', status: '301')
 
+  get '/cycles', to: 'switcher#cycles', as: :cycles
+
+  unless HostingEnvironment.production?
```
duncanjbrown, line 21: Do we need this now that we have the guard in the controller?

spec/spec_helper.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'site_prism'
 require 'simplecov'
+require './app/lib/redis'
```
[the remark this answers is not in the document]
JR-G: Currently, every spec fails without it:
image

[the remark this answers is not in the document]
duncanjbrown: ok let's leave it for now and come back — it should work!

duncanjbrown (approved):

PR #867, opened 2021-08-05
Remove cycle ending soon page and fix banner bug

davidgisbey:
### Context

The code related to cycle ending soon should have been pulled out as part of the deadline banner work, but it was either missed or has found it's way back into the codebase

Additionally, there's bug with the deadline banner. When Find is closed it should not be rendered.

- Remove all code related to the cycle ending soon page
- Don't render the deadline banner when Find is closed as the page already has all the required information


TobyRet (approved):
LGTM

PR #909, opened 2021-08-23
Remove location & provider filters from the sidebar and move into a summary

JR-G:
### Context

Given we will be adding more search filters (visa sponsorship and minimum degree grade), we need to increase the visibility of these new options, and reduce the overall height of the sidebar.

As location and subject options are less likely to change once selected, we can instead summarise the users selection above the page heading, and offer them a link to go back to the landing page to change their selection.

Move the location and provider filters into a summary line above:

image

### Trello card

https://trello.com/c/O1E3RODA

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


JR-G:
Need to set a value for the number of selected subjects that can be displayed on the page. This is currently set to 10
image

spec/features/result_filters/location_and_provider_spec.rb
```diff
@@ -31,39 +32,34 @@
...
       it 'retains the query parameters' do
         expect_page_to_be_displayed_with_query(
           page: results_page,
           expected_query_params: {
-            'fulltime' => 'false',
-            'parttime' => 'false',
-            'hasvacancies' => 'true',
             'l' => '3',
-            'qualifications' => %w[QtsOnly PgdePgceWithQts Other],
-            'senCourses' => 'false',
```
JR-G, line 65: Interesting one here @duncanjbrown... These default params are not in the URL when I manually walk through this and land on the results page for the first time, yet the test is able to pull them from the `current_url`? On my branch they don't appear, hence why I've removed them but behaviour, re params, locally between my branch & master is the same. Any ideas? cc @davidgisbey

app/components/location_subject_filter_component.rb
```diff
@@ -0,0 +1,13 @@
...
+  include ViewHelper
+
+  attr_reader :results
+
+  def initialize(results:)
+    @results = results
+  end
+
+  def subjects_line
+    @results.subjects.map { |subject| "<b>#{subject[:subject_name]}</b>" }.to_sentence(last_word_connector: ' and ').html_safe
```
frankieroberto, line 11: Using Rails’s version of `to_sentence` and the `tag` helper might avoid the need to mark the string as `html_safe`?

```suggestion
    to_sentence(@results.subjects.map { |subject| tag.b(subject[:subject_name) }, last_word_connector: ' and ')
```

app/views/results/index.html.erb
```diff
@@ -1,7 +1,8 @@
 <%= content_for :page_title, @number_of_courses_string %>
 
+<%= render LocationSubjectFilterComponent.new(results: @results_view)%>
+
 <h1 class="govuk-heading-xl" data-qa="heading">
-  <span class="govuk-caption-l">Teacher training courses</span>
   <span data-qa="course-count"><%= "#{@results_view.number_of_courses_string} found" %></span>
```
frankieroberto, line 6: Could we drop this span now (and move the `data-q` up to the `h1` if needed)?

TobyRet (approved):

frankieroberto (approved):
👍  great stuff

PR #928, opened 2021-09-02
Ensure tests pass at all phases of the cycle

davidgisbey:
### Context

As moving between discreet phases of the cycle has now been automated. We need ensure that our tests are Timecop'd. Otherwise when Find closes most of our test suite will break.

I set the time for each test to each discreet phase in the cycle 

image

Then ran all the tests. They all passed except when Find was closed. Where the majority failed as expected.

This means that all the tests that require Timecop to be set to a specific time are. 

I've got with the path of least resistance with this PR and just set a default time for all tests to the `1/1/current_recruitment_cycle_year` rather than randomising the time like we do on Apply. As Find is a much simpler application which isn't interacting with things like RDB/DBD interview dates etc. randomisation felt a bit overkill. 

- Update the queries/factories etc. in Find to use the CycleTimetable.current_year method rather than pulling the current year from the settings file
- Set the default time in specs to be `1/1/current_recruitment_cycle_year`

### Trello card

https://trello.com/c/qMntYngL/3789-audit-tests-in-find-to-see-which-fail-in-each-discreet-phase-of-the-cycle

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


spec/spec_helper.rb
```diff
@@ -95,6 +95,12 @@
   require 'factory_bot'
   config.include FactoryBot::Syntax::Methods
 
+  config.around do |example|
+    Timecop.travel(DateTime.new(RecruitmentCycle.current_year, 1, 1)) do
```
JR-G, line 99: 👍

JR-G (approved):

PR #931, opened 2021-09-02
Return next cycle's courses when changing CycleSwitcher to 'find_reopens'

TobyRet:
### Context

We have been adding new search filters to Find but these rely on data that is only available in courses from the next recruitment cycle.

- When the CycleSwitcher is updated to `find_reopens` the TTAPI now returns courses from the next recruitment cycle.
- Update codebase to obtain the 'current_year' from the CycleTimetable class, rather than Settings

### Guidance to review
- Go to `/cycles` and switch to `find_has_reopened`
- Run a search - you should now see courses from the next recruitment cycle

### Trello card
https://trello.com/c/PmNodHKW/3911-cycleswitcher-add-ability-to-query-next-years-recruitment-cycle-data

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -98,4 +98,4 @@ en:
       description: Candidates can no longer browse courses on Find
     today_is_after_find_opens:
       name: Find has reopened
-      description: Candidates can browse courses on Find
+      description: Candidates can browse courses on Find. Courses returned from the Teacher Traininig API will be from next year's recruitment cycle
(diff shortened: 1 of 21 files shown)


config/locales/en.yml
```diff
@@ -98,4 +98,4 @@ en:
       description: Candidates can no longer browse courses on Find
     today_is_after_find_opens:
       name: Find has reopened
-      description: Candidates can browse courses on Find
+      description: Candidates can browse courses on Find. Courses returned from the Teacher Traininig API will be from next year's recruitment cycle
```
JR-G, line 101: 👍

JR-G (approved):
I've tested this locally and switching to find reopens returned courses from the new cycle

PR #932, opened 2021-09-02
Cache API request made when rendering CoursesController#show 

malcolmbaig:
### Context
Skylight has flagged this as a relatively popular endpoint that could
potentially be made faster. Most of the time in this action is spent
making the API call to TTAPI. 
By setting an expiring cache entry for
each course, we increase the likelihood that popular course data is
served from the cache rather than the API.
### Guidance to review
- We have rack mini profiler on Find now, so viewing a course page on this branch locally should show the API request happening only on the first page load.

### Trello card
https://trello.com/c/qqZRrmQg

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -4,7 +4,8 @@ class CoursesController < ApplicationController
   before_action :render_feedback_component, only: :show
 
   def show
-    @course = Course
+    @course = Rails.cache.fetch(course_cache_key, expires_in: 15.minutes) do
+      Course
       .includes(:subjects)
       .includes(site_statuses: [:site])
       .includes(provider: [:sites])
@@ -13,6 +14,7 @@ class CoursesController < ApplicationController
       .where(provider_code: params[:provider_code])
       .find(params[:course_code])
       .first
+    end
   rescue JsonApiClient::Errors::NotFound
     render template: 'errors/not_found', status: :not_found, formats: [:html]
   end
@@ -30,5 +32,11 @@ class CoursesController < ApplicationController
     redirect_to "#{Settings.apply_base_url}/candidate/apply?providerCode=#{course.provider.provider_code}&courseCode=#{course.course_code}"
   rescue JsonApiClient::Errors::NotFound
     render template: 'errors/not_found', status: :not_found, formats: [:html]
+  end
+
+private
+
+  def course_cache_key
+    "ttapi-course-#{params[:course_code]}-#{params[:provider_code]}"
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/courses_controller.rb
```diff
@@ -31,4 +33,10 @@ def apply
   rescue JsonApiClient::Errors::NotFound
     render template: 'errors/not_found', status: :not_found, formats: [:html]
   end
+
+private
+
+  def course_cache_key
+    "ttapi-course-#{params[:course_code]}-#{params[:provider_code]}"
```
steventux, line 40: Do we sanitise or check these params in any way before using them as cache keys? I wonder how vulnerable this would be to over-stuffing the cache?

steventux (approved):
:+1: Looks good, one question about use of params in cache keys. Happy to discuss.

app/controllers/courses_controller.rb
malcolmbaig: Good point! We don't currently, but in this case the cache key doesn't get stored unless it's a valid course.

app/controllers/courses_controller.rb
```diff
@@ -4,7 +4,8 @@ class CoursesController < ApplicationController
   before_action :render_feedback_component, only: :show
 
   def show
-    @course = Course
+    @course = Rails.cache.fetch(course_cache_key, expires_in: 15.minutes) do
```
malcolmbaig, line 7: This means that any changes to course details will take (at most) 15 minutes to update on Find if the course is in the cache. Is that ok? @davidgisbey @duncanjbrown

config/environments/test.rb
```diff
@@ -20,6 +20,9 @@
     'Cache-Control' => 'public, max-age=3600',
   }
 
+  # Ensure that our tests are accounting for caching behaviour
+  config.cache_store = :memory_store
```
stevehook, line 24: I was just wondering how much memory the cache would take up, how big it might get if we cache a lot of these kinds of request. Default size for the cache is 32MB according to the docs so I guess it's not going to get excessive.

malcolmbaig:
Speaking to Publish/Register, this approach will kick up some usability problems if a provider wants to change/correct some course details quickly in Publish. We could reduce the cache time here, but an alternative suggested by @gpeng in the [discussion](https://ukgovernmentdfe.slack.com/archives/CAHAJFR0X/p1630664825188600) is to cache the course details TTAPI-side and do invalidation there when the course is updated. 

I'm going to spike this and see how the load test results respond.

PR #940, opened 2021-09-08
Return 'next year' if cycle switcher has been set to 'find has reopened'

TobyRet:
### Context
[Following a recent change](https://github.com/DFE-Digital/find-teacher-training/pull/928/commits/6522cf5e99e066be9085cab878a461d2a879a727) when a user switches the cycle to 'find has reopened', 2022 courses are no longer displayed. To be fair the original implementation was substandard and this new implementation is simpler (credit: @davidgisbey)

- Return 2022 courses from the TTAPI when the cycle switcher is set to 'find has reopened'
- Remove old logic

### Guidance to review
Visit `/cycles` and set the cycle switcher to 'find has reopened'. Run a search and you should see that 2022 courses are returned in the results

### Trello card
https://trello.com/c/rwSlAEhY/3925-fix-cycle-switcher-find-reopens

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


JR-G (approved):
tested locally and it's returning 2022 courses and using the old switcher

PR #959, opened 2021-09-15
Configure assets domain for serving static assets

vigneshmsft:
### Context

At present the static assets are being served directly from the rails app server.
We can reduce the number of requests served by the app by serving assets from a different cdn-route instance.

Set RAILS_ASSETS_HOST for different envs in KeyVault secrets
Terraform config for binding assets domain to web app
Set `asset_host` rails property to the assets domain

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/terraform/modules/paas/variables.tf
+++ b/terraform/modules/paas/variables.tf
@@ -47,6 +47,14 @@ locals {
     prod    = ["www", "www2"]
     review  = [local.app_name_suffix]
   }
+  assets_host_names = {
+    qa        = "qa-assets"
+    staging   = "staging-assets"
+    sandbox   = "sandbox-assets"
+    prod      = "assets"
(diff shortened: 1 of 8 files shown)


terraform/modules/paas/variables.tf
```diff
@@ -47,6 +47,15 @@ locals {
     prod    = ["www", "www2"]
     review  = [local.app_name_suffix]
   }
+  assets_host_names = {
+    qa        = "qa-assets"
+    staging   = "staging-assets"
+    sandbox   = "sandbox-assets"
+    loadtest = "assets"
```
malcolmbaig, line 54: doest this need to be `qa-assets`, `loadtest-assets`, or similar?

saliceti (approved):

PR #970, opened 2021-09-17
Update subject params to use rails subject codes

davidgisbey:
### Context

The subject filters in Find populate the query string with subject codes. These subject codes are the so-called Csharp subject codes, dating from when Find was a C# app before it was rewritten in Rails.

We would like to remove the Csharp codes and replace them with so-called Rails (ie TTAPI-standard) codes so that every service is using the same codes. Apart from being the right thing to do, this also simplifies our reporting and simplifies Find itself.

**However**, we do need to continue to support the old subject codes, because we want to continue to support users who have bookmarked search result pages. (A quick investigation with @misaka found a few hundred requests lacking `Referer`s a month).

- Change Find subject filters to use the same subject codes as TTAPI does (it doesn't rn)
- Continue to support legacy codes using the `CsharpRailsSubjectConversionHelper`

### Guidance to review

Have i missed anything?

### Trello card

https://trello.com/c/MxJ6HtX4/3952-remove-the-csharp-to-rails-subject-conversion-logic-in-the-subject-filters

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/result_filters/subject_controller.rb
+++ b/app/controllers/result_filters/subject_controller.rb
@@ -16,7 +16,7 @@ module ResultFilters
     end
 
     def create
-      if params[:rails_subjects].blank? && (params[:senCourses].blank? || params[:senCourses] == 'false')
+      if params[:subject_codes].blank? && (params[:senCourses].blank? || params[:senCourses] == 'false')
         flash[:error] = [I18n.t('subject_filter.errors.no_option')]
 
         if flash[:start_wizard]
@@ -25,14 +25,16 @@ module ResultFilters
           redirect_to subject_path(filter_params)
         end
       else
-        redirect_to results_path(filter_params.merge(rails_subjects: params[:rails_subjects]))
+        redirect_to results_path(filter_params.merge(subject_codes: params[:subject_codes]))
       end
     end
 
   private
 
     def convert_csharp_params_to_rails
-      params['rails_subjects'] = convert_csharp_subject_id_params_to_subject_code if convert_csharp_subject_id_params_to_subject_code.present?
+      if params['subject_codes'].blank? && convert_csharp_subject_id_params_to_subject_code.present?
+        request.query_parameters['subject_codes'] = convert_csharp_subject_id_params_to_subject_code
+      end
     end
 
     def build_subject_areas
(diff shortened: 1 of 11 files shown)


app/controllers/result_filters/subject_controller.rb
```diff
@@ -16,7 +16,7 @@ def start
     end
 
     def create
-      if params[:subjects].blank? && (params[:senCourses].blank? || params[:senCourses] == 'false')
+      if params[:subject_codes].blank? && (params[:senCourses].blank? || params[:senCourses] == 'false')
```
TobyRet, line 19: Very minor nit pick but wondering if we should be using the `subject_code` rather than `subject_codes` so we have some thing like:
`/results?l=2&subject_code%5B%5D=00&subject_code%5B%5D=01&subject_code%5B%5D=02`

There is also a larger question about whether Find params should be hyphenated rather than snake-cased

app/helpers/result_filters/subject_helper.rb
```diff
@@ -4,17 +4,17 @@ module SubjectHelper
     include CsharpRailsSubjectConversionHelper
```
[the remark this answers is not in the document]
TobyRet: Actually I see there are some comments in there already by may need reviewing / updating - particularly this one
https://github.com/DFE-Digital/find-teacher-training/blob/master/app/helpers/csharp_rails_subject_conversion_helper.rb#L1

TobyRet (commented):
Couple of minor comments otherwise LGTM

app/controllers/result_filters/subject_controller.rb
davidgisbey: So the problem is though that it munges it into an array in the controllers so it's always going to be dramatically incorrect somewhere. Feels like this this might be the lesser evil otherwise you're iterating over `subject_code` wdyt?

app/helpers/result_filters/subject_helper.rb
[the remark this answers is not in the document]
davidgisbey: have updated the comment

app/controllers/result_filters/subject_controller.rb
TobyRet: Ah right - np

TobyRet (approved):

PR #976, opened 2021-09-21
Add temporary caching of course API requests

malcolmbaig:
### Context
We can't cache course details for any significant length of time in
Find, because there's a requirement for Providers to be able to quickly
update course details and issue corrections. However, we probably should
cache these API calls anyway during the opening week of the cycle, to
help the service handle the expected level of load.

Do this via the following:

- Add a feature flag for this caching behaviour
- Use Faraday middleware to write to a cache when making requests in the
  Course resource
- Give cache entries written here a 10 minute expiry
- Make the Course connection use a valid cache object if the feature
  flag is on, otherwise use a null cache

This will serve requests for the same url, within the expiry, from the
cache.

Alongside these changes, we configure as follows:

- Clear the cache between tests
- Switch caching on in development, test, and qa
- Switch caching off in production (for now)
- ~Set production explicitly to a null store for now, changing this to a
  Redis store once an instance is configured.~

### Guidance to review
- This is enabled in development, and can be tested locally by navigating to search results and course pages. After the first visit, subsequent visits should skip the API call to TTAPI (visible in rack mini profiler). Change the value of `Course::TTAPI_CALLS_EXPIRY` to something lower (eg—10.seconds) to test the expiration.
- Note that both search queries and individual course detail pages have their TTAPI response data cached.
- What do we think of the overall approach here, namely switching this on temporarily before the spike?

### Trello card
https://trello.com/c/i4k7rJpC
### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -15,8 +15,6 @@ class ResultsController < ApplicationController
     begin
       @courses = @results_view.courses.all
       @number_of_courses_string = @results_view.number_of_courses_string
-
-      expires_in 10.minutes
     rescue JsonApiClient::Errors::ClientError
       render template: 'errors/unprocessable_entity', status: :unprocessable_entity
     end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -20,6 +20,9 @@ Rails.application.configure do
     'Cache-Control' => 'public, max-age=3600',
   }
 
+  # Ensure that our tests are accounting for caching behaviour
+  config.cache_store = :memory_store
+
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,4 +1,16 @@
 class Course < Base
+  TTAPI_CALLS_EXPIRY = 10.minutes
+
+  connection do |conn|
+    conn.faraday.response(:caching, write_options: { expires_in: TTAPI_CALLS_EXPIRY }) do
+      if Settings.feature_flags.cache_courses
+        Rails.cache
+      else
+        ActiveSupport::Cache::NullStore.new
+      end
+    end
+  end
+
   belongs_to :recruitment_cycle, through: :provider, param: :recruitment_cycle_year
   belongs_to :provider, param: :provider_code, shallow_path: true
   has_many :site_statuses
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -24,3 +24,4 @@ feature_flags:
   new_filters: false
   maintenance_mode: false
   maintenance_banner: false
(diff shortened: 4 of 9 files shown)


app/controllers/results_controller.rb
```diff
@@ -13,8 +13,6 @@ def index
     begin
       @courses = @results_view.courses.all
       @number_of_courses_string = @results_view.number_of_courses_string
-
-      expires_in 10.minutes
```
malcolmbaig, line 17: To keep the caching behaviour as simple as possible, remove the existing client-side caching we're doing here. This only works in a subset of browsers anyway (eg—Chrome ignores `max-age` a lot of the time).

config/environments/production.rb
```diff
@@ -50,7 +50,7 @@
   config.rails_semantic_logger.add_file_appender = false
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store, { url: ENV.fetch('REDIS_CACHE_URL') }
```
malcolmbaig, line 53: @vigneshmsft Is this correct?

config/environments/test.rb
```diff
@@ -20,6 +20,9 @@
     'Cache-Control' => 'public, max-age=3600',
   }
 
+  # Ensure that our tests are accounting for caching behaviour
+  config.cache_store = :memory_store
```
duncanjbrown, line 24: Do we have facilities for testing this? eg hit a fake API endpoint, render the page, change what's in the API, check the page is unchanged, clear the cache, check the page is changed?

config/environments/production.rb
```diff
@@ -50,7 +50,7 @@
   config.rails_semantic_logger.add_file_appender = false
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store, { url: RedisService.redis_url }
```
malcolmbaig, line 53: Needs reviewing - maybe need a separate Redis in prod?

config/environments/test.rb
malcolmbaig: Will put something together 👍

malcolmbaig: @duncanjbrown added a spec, wdyt?

steventux (approved):
Looks good, given the complexities of caching upstream I think this is a good approach. 

Not for this PR but it might be worth iterating this so we have a manual way to invalidate the cached results (perhaps even specific to the search params) as this might allow us longer expiry times.

duncanjbrown (approved):

config/environments/production.rb
malcolmbaig: Now pointing to a cache instance in master, thanks to @vigneshmsft

PR #978, opened 2021-09-21
Add maintenance page and flag for when find is down

davidgisbey:
### Context

We're going to have to bring Find down for a set period of time while we upgrade the TTAPI DB. There may well be a few minutes where we're unable to hit the TTAPI so we're going to pop the service into maintenance mode for 30 mins.

- Add maintenance_mode feature flag
- Add a maintenance mode page 
- Redirect to the page when the flag is active

### Guidance to review

Does this seem about right time wise? Not sure how long the db upgrade will take. Could make it an hour to be safe?

This will have to be manually deployed before we swap the DBs. Is that an issue?

image

Add banner 

image

🤢

### Trello card

https://trello.com/c/WPcZ8odw/3970-add-maintenance-page-for-find-for-when-the-ttapi-is-down

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :store_request_id
   before_action :assign_sentry_contexts
+  before_action :redirect_to_maintanance_page_if_flag_is_active
   before_action :redirect_to_cycle_has_ended_if_find_is_down
 
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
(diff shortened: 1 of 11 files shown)


defong (dismissed):

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :store_request_id
   before_action :assign_sentry_contexts
+  before_action :redirect_to_maintanance_page_if_flag_is_active
```
defong, line 8:
```suggestion
  before_action :redirect_to_maintenance_page_if_flag_is_active
```

malcolmbaig (approved):

PR #988, opened 2021-09-24
[2261] Stop logging Bigquery job

saliceti:
### Context
Logit is expensive and most of the logs are because of the BigQuery job. They don't add any value since we use BigQuery as an event log anyway.

Turn off logging for the BigQuery job. Same as teaching vacancies: https://github.com/DFE-Digital/teaching-vacancies/pull/3986/files

### Guidance to review
Check the BigQuery job doesn't emit any log

### Trello card
https://trello.com/c/7uDtQqB3

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/job/send_event_to_big_query_job.rb
+++ b/app/job/send_event_to_big_query_job.rb
@@ -1,4 +1,6 @@
 class SendEventToBigQueryJob < ApplicationJob
+  self.logger = ActiveSupport::TaggedLogging.new(Logger.new(IO::NULL))
+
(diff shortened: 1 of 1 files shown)


misaka (approved):

duncanjbrown (approved):
Looks good! Worth adding a note in the commit message about why we need to wrap in `TaggedLogging`? (needs to implement the `tagged` method to work with `SemanticLogger`)

PR #996, opened 2021-09-28
Disable cycle switcher on Find for the load test environment

davidgisbey:
### Context

This is disabled for Prod so it should also be for our load testing env.

- Disable the cycle switcher in the load testing env

### Trello card

https://trello.com/c/YW5sKe0d/4002-dont-use-redis-on-prod-load-testing-in-find

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/switcher_controller.rb
+++ b/app/controllers/switcher_controller.rb
@@ -14,6 +14,8 @@ class SwitcherController < ApplicationController
 private
 
   def redirect_to_homepage_if_in_production
-    redirect_to root_path if Rails.env.production?
+    # rubocop:disable Rails/UnknownEnv
+    redirect_to root_path if Rails.env.production? || Rails.env.loadtest?
+    # rubocop:enable Rails/UnknownEnv
   end
 end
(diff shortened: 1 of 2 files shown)


app/lib/hosting_environment.rb
```diff
@@ -2,4 +2,8 @@ module HostingEnvironment
   def self.production?
     Rails.env.production?
   end
+
+  def self.loadtest?
+    Rails.env.loadtest?
```
duncanjbrown, line 7: I don't think we have a `loadtest.rb` in `environments`?

malcolmbaig: We have one on Find's load-test branch!

malcolmbaig (approved):

app/lib/hosting_environment.rb
duncanjbrown: Aha!

malcolmbaig:
re: linter—I'd either use a rubocop disable comment or do a string comparison of the env instead.

PR #1005, opened 2021-09-29
Use net_http_persistent for better performance when making API calls

vigneshmsft:
### Context

Calls to the Google API are quite slow, according to mini profiler:

image

We can improve this by using persistent HTTP connections: 

image

- Remove HTTParty gem
- Add persistent http gem
- Use Faraday for connecting to Google API, with net_http_persistent as the adaptor
- Use a simlar approach for making all API calls to TTAPI

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,8 @@ gem 'pkg-config', '~> 1.4.6'
 # Decorate logic to keep it of the views and helper methods
 gem 'draper'
 
-# Keep-Alive a http connection
+# HTTP client library
+gem 'faraday'
 gem 'net-http-persistent'
 
 # Semantic Logger makes logs pretty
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -10,8 +10,8 @@ gem 'pkg-config', '~> 1.4.6'
 # Decorate logic to keep it of the views and helper methods
 gem 'draper'
 
-# http client
-gem 'httparty'
+# Keep-Alive a http connection
+gem 'net-http-persistent'
```
malcolmbaig, line 14:
```suggestion
gem 'net-http-persistent'

# HTTP client library
gem 'faraday'
```

Worth specifying this explicitly as a dependency so its clear that this is the http lib of choice for this project (so we don't see the reintroduction of httparty or other).

malcolmbaig (commented):
One minor comment but otherwise LG 👍🏼

duncanjbrown (approved):

misaka:
😍

malcolmbaig (approved):

PR #1008, opened 2021-09-30
Feature flag provider autocomplete

malcolmbaig:
### Context
The provider autocomplete on the start page sends requests to TTAPI as
the user types. This gets a bit heavy-going when traffic spikes at start
of cycle.

We're planning on reimplementing it so it uses a locally-stored
providers list for its suggestions. Until that's done, we want to be
able to switch the autocomplete off during peak load.

Wrap the container div for the autocomplete in a feature flag check
Return early from the autocomplete js if this div isn't found on the page

### Guidance to review
- Can be tested locally by toggling the value of the feature flag and typing a provider name into the third input on the start page

### Trello card
https://trello.com/c/B3FlzE6Z
### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/views/result_filters/location/_form.html.erb
+++ b/app/views/result_filters/location/_form.html.erb
@@ -101,12 +101,15 @@
                   </span>
                   <% end %>
                 <% end %>
-                <%= form.text_field :query,
-                                    id: 'provider',
-                                    value: params[:query],
-                                    class: 'govuk-input',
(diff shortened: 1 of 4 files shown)


TobyRet (approved):
Tested locally and LGTM

davidgisbey (approved):

PR #1010, opened 2021-09-30
Update QTS-only description to reflect UK leaving the EU

frankieroberto:
The QTS-only details component contained some content written pre-Brexit.

This brings it inline with the other qualification statuses in advising candidates to check qualification requirements for overseas teaching.

# Before

Screenshot 2021-09-30 at 14 19 17

# After

Screenshot 2021-09-30 at 14 20 01

### DIFF
--- a/app/views/courses/qualification/_qts.html.erb
+++ b/app/views/courses/qualification/_qts.html.erb
@@ -3,6 +3,6 @@
     Qualified teacher status (QTS) allows you to teach in state schools in England and may also allow you to teach in other parts of the UK.
   </p>
   <p class="govuk-body">
-    It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you&rsquo;d like to teach in.
+    It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
   </p>
 <% end %>
--- a/app/views/courses/qualification/_qts.html.erb
+++ b/app/views/courses/qualification/_qts.html.erb
@@ -3,9 +3,6 @@
     Qualified teacher status (QTS) allows you to teach in state schools in England and may also allow you to teach in other parts of the UK.
   </p>
   <p class="govuk-body">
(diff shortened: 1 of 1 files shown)


app/views/courses/qualification/_qts.html.erb
```diff
@@ -3,9 +3,6 @@
     Qualified teacher status (QTS) allows you to teach in state schools in England and may also allow you to teach in other parts of the UK.
   </p>
   <p class="govuk-body">
-    It may also allow you to teach in <%= govuk_link_to 'the EU and EEA', 'https://www.gov.uk/eu-eea' %>, though this could change after 2020.
-  </p>
-  <p class="govuk-body">
-    If you&rsquo;re planning to teach overseas, you should always check what qualifications are needed in the country you&rsquo;d like to teach in.
+    It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you&rsquo;d like to teach in.
```
duncanjbrown, line 6:
```suggestion
    It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
```

Any reason not to do this?

duncanjbrown (approved):
:+1:

Anyone know why we're manually HTML-encoding our own hardcoded text? I would expect to see that in user-submitted content, and for it to be done mechanically

app/views/courses/qualification/_qts.html.erb
frankieroberto: Nope. I just copied and pasted from the other files... 😄

frankieroberto:
> Anyone know why we're manually HTML-encoding our own hardcoded text? I would expect to see that in user-submitted content, and for it to be done mechanically

@duncanjbrown good point – I've updated all these here: https://github.com/DFE-Digital/find-teacher-training/pull/1012

I've left the `&nbsp;` entities (non-breaking spaces), as whilst it's also possible to directly encode these, they look identical to normal spaces, and so they could easily get edited out by mistake.

PR #1011, opened 2021-10-01
Cache location suggestions

malcolmbaig:
### Context
The [location suggestions endpoint](https://www.skylight.io/app/applications/fjR4NhKjEKlK/recent/6h/endpoints/LocationSuggestionsController%23index?responseType=json) gets hit at fairly high volumes as a user types a city, town or post code on the start page. Temporarily caching the results of the endpoint may help improve performance and avoid repeatedly sending the same input queries to the Google API in a short space of time.

- Use debounce instead of throttle in the autocomplete, to improve the quality of search queries sent to the server
- Cache valid results with an expiry, ignoring errors from the Google API

### Guidance to review
Can be tested locally by navigating to the start page and searching by city, town or post code.

Edit -
Via @duncanjbrown, a demo of the difference between throttle and debounce http://demo.nimius.net/debounce_throttle/

### Trello card
https://trello.com/c/FtWwPETn

### Checklist

- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -1,20 +1,33 @@
 class LocationSuggestion
+  CACHE_EXPIRY = 30.minutes
+
   class << self
     def suggest(input)
       query = build_query(input)
 
+      cached_result = Rails.cache.read(location_query_cache_key(input))
+      return cached_result if cached_result.present?
+
       response = connection.get("#{Settings.google.places_api_path}?#{query.to_query}").body
 
-      if response['predictions'].present?
-        response['predictions']
-          .map(&format_prediction)
-          .take(5)
-      elsif response['error_message'].present?
+      if response['error_message'].present?
         Sentry.capture_message(message: response['error_message'])
+        return
       end
+
+      result =
+        if response['predictions'].present?
+          response['predictions'].map(&format_prediction).take(5)
+        end
+      Rails.cache.write(location_query_cache_key(input), result, expires_in: CACHE_EXPIRY)
+      result
     end
 
   private
+
+    def location_query_cache_key(input)
+      "location-query-#{input.upcase}"
+    end
 
     def connection
       Faraday.new(Settings.google.places_api_host) do |f|
--- a/app/services/location_suggestion.rb
+++ b/app/services/location_suggestion.rb
@@ -4,26 +4,34 @@ class LocationSuggestion
   class << self
     def suggest(input)
       query = build_query(input)
-
       cached_result = Rails.cache.read(location_query_cache_key(input))
-      return cached_result if cached_result.present?
+      return cached_result unless cached_result.nil?
 
-      response = connection.get("#{Settings.google.places_api_path}?#{query.to_query}").body
+      response = connection.get("#{Settings.google.places_api_path}?#{query.to_query}")
 
-      if response['error_message'].present?
-        Sentry.capture_message(message: response['error_message'])
+      if do_not_cache?(response)
+        report_error(response)
         return
       end
 
-      result =
-        if response['predictions'].present?
-          response['predictions'].map(&format_prediction).take(5)
-        end
+      result = response.body['predictions'].map(&format_prediction).take(5)
       Rails.cache.write(location_query_cache_key(input), result, expires_in: CACHE_EXPIRY)
       result
     end
 
   private
+
+    def do_not_cache?(response)
+      response.status != 200 ||
+        response.body['error_message'].present? ||
+        response.body['predictions'].blank?
+    end
+
+    def report_error(response)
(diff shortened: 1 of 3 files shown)


malcolmbaig:
@TobyRet fyi, this may possibly influence how you approach reworking the provider autocomplete.

app/services/location_suggestion.rb
```diff
@@ -1,21 +1,34 @@
 class LocationSuggestion
+  CACHE_EXPIRY = 30.minutes
+
   class << self
     def suggest(input)
       query = build_query(input)
 
+      cached_result = Rails.cache.read(location_query_cache_key(input))
+      return cached_result if cached_result.present?
```
duncanjbrown, line 9: Would this logic work using `cache.fetch` with a block? https://apidock.com/rails/ActiveSupport/Cache/Store/fetch. If there's an error we'll end up with `nil` in the cache anyway

app/webpacker/scripts/autocomplete.js
```diff
@@ -48,7 +48,7 @@ const initAutocomplete = ({element, input, path, selectNameAndCode}) => {
         name: $input.name,
         defaultValue: $input.value,
         minLength: 3,
-        source: throttle(request(path), 2000),
+        source: debounce(request(path), 900),
```
duncanjbrown, line 51: :+1:

spec/services/location_suggestion_spec.rb
```diff
@@ -61,6 +65,38 @@ def stub_query(status: 200, predictions: {}, error_message: nil)
...
+        query_stub
+      end
+
+      it 'caches suggestions for a period of time' do
+        allow(Rails.cache).to receive(:read).and_call_original
+        allow(Rails.cache).to receive(:write).and_call_original
+
+        result = suggest!
+        expect(result.count).to eq 3
+        expect(Rails.cache).to have_received(:read).once
```
duncanjbrown, line 79: Do we need to assert on these calls? Whether the request gets made or not feels like the salient thing

app/services/location_suggestion.rb
```diff
@@ -1,21 +1,34 @@
...
+          response['predictions'].map(&format_prediction).take(5)
+        end
+      Rails.cache.write(location_query_cache_key(input), result, expires_in: CACHE_EXPIRY)
+      result
     end
 
   private
 
+    def location_query_cache_key(input)
+      "location-query #{input.upcase}"
```
duncanjbrown, line 29: Maybe no space in the cache key? (probably doesn't matter, but feels like it could)

duncanjbrown (commented):
Looks great, nits only!

app/services/location_suggestion.rb
malcolmbaig: I wanted to avoid writing an entry in the cache if there's an error with the API call, but still write nil if there are genuinely no results for the query. Couldn't see how to do that with fetch.

duncanjbrown: Ah ok. So set the key, but store nil against it? And not set the key if there was an error

stevehook: Would it make sense to store `[]` in the cache if there are no results? Otherwise you can't really distinguish between the 'no result' and 'new key' cases?

stevehook (approved):
Looks good to me, one little question. Worked for me locally (once I realised that I had to change https://github.com/DFE-Digital/find-teacher-training/blob/master/config/environments/development.rb#L30)

app/services/location_suggestion.rb
duncanjbrown: That appeals

malcolmbaig: @stevehook good point! It's actually an empty array that gets stored, so what needs changing here is the guard clause. Will update.

malcolmbaig:
> Looks good to me, one little question. Worked for me locally (once I realised that I had to change https://github.com/DFE-Digital/find-teacher-training/blob/master/config/environments/development.rb#L30)

@stevehook can also use `rails dev:cache` on the cli to toggle between modes!

app/services/location_suggestion.rb
malcolmbaig: @stevehook @duncanjbrown fyi, I've changed my approach here. It's genuinely quite difficult to get an empty result set without typing garbage into the query field. On that basis, I think this class should:
- Write a cache entry if it gets predictions back
- Do nothing in the following scenarios:
  - Server error
  - Error message in the response body (eg—API key invalid, reported in a response with status 200 😞 )
  - Empty result set

WDYT?

duncanjbrown: Sounds good!

duncanjbrown (approved):

PR #1014, opened 2021-10-04
Remove bursary and scholarship information and add placeholder guidance

davidgisbey:
### Context

We don't have information regarding bursaries of scholarships at the moment. We also need to pull the SKE related stuff out as we're not sure which will have it.

- Put all of the bursary/scholarship/SKE content behind a flag 

- Subject page content

|Before|After|
|-----------|-----------|
|image|image|

- Subject page filters

|Before|After|
|-----------|-----------|
|image|image|

- Results page 

|Before|After|
|-----------|-----------|
|image|image|

- Course page

|Before|After|
|-----------|-----------|
|image|image|

### Guidance to review

Have i got everything bursary/scholarship/ske related? The doc was missing a few things. Might be worth having a bit of a manual test with courses in different states.

### Trello card

https://trello.com/c/DPLthMSD/4003-dev-changes-to-find-when-the-new-cycle-is-open-but-funding-has-not-been-anncounced

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -26,4 +26,4 @@ feature_flags:
   maintenance_banner: false
   cache_courses: false
   provider_autocomplete: true
-  bursaries_and_scholarships_announced: false
+  bursaries_and_scholarships_announced: true
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -26,3 +26,4 @@ feature_flags:
   maintenance_banner: false
   cache_courses: false
   provider_autocomplete: true
+  bursaries_and_scholarships_announced: false
--- a/app/components/courses/search_result_component.html.erb
+++ b/app/components/courses/search_result_component.html.erb
@@ -26,8 +26,10 @@
       <% end %>
     <% end %>
 
-    <dt class="app-description-list__label">Financial support</dt>
-    <dd data-qa="course__funding_options"><%= course.decorate.funding_option %></dd>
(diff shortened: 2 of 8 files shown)


config/settings.yml
```diff
@@ -26,3 +26,4 @@ feature_flags:
   maintenance_banner: false
   cache_courses: false
   provider_autocomplete: true
+  bursaries_and_scholarships_announced: true
```
TobyRet, line 29: Shouldn this be `false`?

TobyRet (commented):
LGTM - tested locally and could not find any other code that should be feature flagged. Just one question about setting the feature flag value
https://github.com/DFE-Digital/find-teacher-training/pull/1014#discussion_r721330225

config/settings.yml
davidgisbey: So it will be tonight at midnight, but it still needs to show for the rest of the day.

After i merge this i'll put together the PR that needs to be merged at midnight

TobyRet: Makes sense - not sure if its worth turning on in QA or Staging so that people can review? ...Or should we direct to the review app?

davidgisbey: yea good shout i'll switch it off in QA

TobyRet (approved):

PR #1016, opened 2021-10-04
Fix bug with subjects in filters

davidgisbey:
### Context

Fix issue with subjects not being retained when filters are applied

- Update the hidden fields to use the new subject_codes field rather than the deprecated params

### Trello
https://trello.com/c/8sXFu9UP/4014-find-send-filter-bug

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


TobyRet (approved):
Tested and LGTM - just needs a spec but approving now

PR #1029, opened 2021-10-05
[DO NOT MERGE] - Cache providers in find

TobyRet:
### Context

`ProviderSuggestionsController#index` in Find is relatively slow (95th percentile response >200ms). It queries TTAPI with whatever the candidate types into the provider text field.

We have <2000 providers. We can try caching the data needed by this endpoint so that we don't need to make as many calls to TTAPI.

- Create job to sync provider names and codes in Redis
- Update the provider autocomplete to use the cached providers 
- Add feature flag

### Guidance to review
- NOTE - this PR cannot be merged until a Clock instance is set up on infra - this will be responsible for calling the sync job every hour
- In `development.local.yaml` set `cache_providers` to `true`
- In the rails console run `SyncAllProviders.call` (Make you have TTAPI running before you do this)

### Trello card
https://trello.com/c/dcKteB0b/3917-cache-providers-in-find

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


spec/services/teacher_training_public_api/sync_all_providers_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe TeacherTrainingPublicAPI::SyncAllProviders do
```
TobyRet, line 3: This is a unit test - wonder if we need an integration spec instead / as well 🤔

PR #1032, opened 2021-10-05
Add `subject_codes` to list of array query params

stevehook:
### Context

We have seen a few errors flagged today in production that are caused by invalid `subject_codes` query params:

https://sentry.io/organizations/dfe-bat/issues/2686958196/?project=1780060

- Ensure that hash values for `subject_codes` are converted to arrays.

### Guidance to review

- I don't fully understand how we are getting the invalid params. Find itself doesn't seem to generate them in my testing. I don't think they are legacy parameters either because `subject_codes` is quite new - that at least explains why we are starting to see this now rather than earlier.

### Trello card

https://trello.com/c/NLdQD0Q6/4021-fix-errors-when-find-search-receives-invalid-subjectcodes

### Checklist

- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/convert_deprecated_csharp_parameters_service.rb
+++ b/app/services/convert_deprecated_csharp_parameters_service.rb
@@ -1,7 +1,7 @@
 class ConvertDeprecatedCsharpParametersService
   def call(parameters:)
     boolean_parameter_names = %w[fulltime hasvacancies parttime senCourses]
-    array_parameter_names = %w[qualifications subjects]
+    array_parameter_names = %w[qualifications subjects subject_codes]
 
     have_legacy_params = false
     params_hash = parameters
--- a/spec/services/convert_deprecated_csharp_parameters_service_spec.rb
+++ b/spec/services/convert_deprecated_csharp_parameters_service_spec.rb
@@ -13,6 +13,7 @@ RSpec.describe ConvertDeprecatedCsharpParametersService do
         'senCourses' => 'True',
         'qualifications' => 'QtsOnly,PgdePgceWithQts,Other',
         'subjects' => '1,2,3',
+        'subject_codes' => { '0' => '00' },
       }
     end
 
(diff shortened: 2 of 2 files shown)


duncanjbrown:
This works... but it's super weird!

It might be related to Facebook? Most (but not all) UAs are explicitly FB, tho all (from casual survey in bigquery) have a `fbclid` param, which FB sticks on outbound links. Missing `referer` implies traffic from a mobile app rather than the web.

It's not just a link shared on FB: the payloads vary on the other fields (tho not on this one). And anyway it would be weird if everyone suddenly updated their bookmarks or whatever with an invalid `subject_codes` bit since we introduced that param a couple of weeks ago.

So it sounds absurd, but it feels a bit like someone is somehow semi-automatically (and wrongly) minting Find search links on Facebook?! (Possibly distant colleagues?)

All for fixing it in preprocessing like this. But maybe we should move the code that deals with this out from the unrelated C# code?

stevehook:
> This works... but it's super weird!
> 
> It might be related to Facebook? Most (but not all) UAs are explicitly FB, tho all (from casual survey in bigquery) have a `fbclid` param, which FB sticks on outbound links. Missing `referer` implies traffic from a mobile app rather than the web.
> 
> It's not just a link shared on FB: the payloads vary on the other fields (tho not on this one). And anyway it would be weird if everyone suddenly updated their bookmarks or whatever with an invalid `subject_codes` bit since we introduced that param a couple of weeks ago.
> 
> So it sounds absurd, but it feels a bit like someone is somehow semi-automatically (and wrongly) minting Find search links on Facebook?! (Possibly distant colleagues?)
> 
> All for fixing it in preprocessing like this. But maybe we should move the code that deals with this out from the unrelated C# code?

Thanks @duncanjbrown makes sense to move this out of the legacy C# middleware.

I had similar thoughts about a Facebook page that somebody has created that provides some kind of interface to Find. All of the the cases I looked at had either a `fbclid` param or came from a Facebook browser or had messenger.com as their referrer. I also saw invalid values for other array params like `qualifications` but we already handle that one.

PR #1070, opened 2021-10-25
Update location content

JR-G:
### Context

A [support ticket](https://trello.com/c/pZatjybt/859-find-page-not-showing-location) was raised by a provider on how we render the location information on Find, when a candidate searches with 'Across England'.

Previously we would render the providers address, this PR introduces logic to conditionally render either their site address (if they only have a single site) or the count of sites (if they have more than one).

- Added a new `_sites.html.erb` partial
- Update `index.html.erb` so that it takes in a `ResultsView` object

The results page will now display location information in either of the below formats:
image

### Guidance to review

I'm not 100% sure if passing ResultsView into the `SearchResultComponent` is the best approach – I ended up doing this because the `search_result_component_spec.rb` tests were failing due to it calling methods like `sites_count` on a nil class. Happy to discuss this – I've also not touched `_non_university.html.erb` or `_university.html.erb` until the approach has been agreed but I guess these could also be updated now that results_view is accessible. 

### Trello card

https://trello.com/c/pZatjybt/859-find-page-not-showing-location

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/spec/features/courses/results_spec.rb
+++ b/spec/features/courses/results_spec.rb
@@ -46,7 +46,8 @@ describe 'Search results', type: :feature do
         expect(first_course.qualification.text).to include('PGCE with QTS')
         expect(first_course.study_mode.text).to eq('Full time')
         expect(first_course.funding_options.text).to eq('Student finance if you’re eligible')
-        expect(first_course.main_address.text).to eq('Poole SCITT, Ad Astra Infant School, Sherborn Crescent, Poole, Dorset, BH17 8AP')
+        expect(first_course.main_address_type.text).to eq('Number of locations')
+        expect(first_course.main_address.text).to eq('4')
         expect(first_course).not_to have_show_vacancies
       end
     end


app/components/courses/search_result_component.html.erb
```diff
@@ -36,10 +36,7 @@
     <dd data-qa="course__visa_sponsorship"><%= visa_sponsorship_status %></dd>
 
     <% unless filtered_by_location? %>
-      <dt class="app-description-list__label">Locations</dt>
-      <dd data-qa="course__main_address">
-        <%= helpers.smart_quotes(course.provider.decorate.short_address) %>
-      </dd>
+        <%= render partial: 'results/sites', locals: { course: course, results_view: results_view } %>
```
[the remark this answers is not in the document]
TobyRet: I _think_  'results/location' is no longer used. We used to show a GoogleMap on the LHS with a pin that representing the candidates location query - 
I removed and ran the specs and it was green so should be removed. 
'Sites' still makes sense to me however, as this is the name of the resource we request from TTAPI

app/components/courses/search_result_component.rb
```diff
@@ -12,10 +12,11 @@ def has_sites?
       @has_sites
     end
 
-    def initialize(course:, filtered_by_location: false, has_sites: false)
+    def initialize(course:, results_view:)
       @course = course.decorate
-      @filtered_by_location = filtered_by_location
-      @has_sites = has_sites
+      @filtered_by_location = results_view.location_filter?
+      @has_sites = results_view.has_sites?(course)
```
TobyRet, line 18: I don't think `@has_sites` needs to be declared as an instance var?

Perhaps just define `#has_sites` as follows 

```
 def has_sites?
     @results_view.has_sites?(@course)
  end
```

TobyRet: Same deal for @filtered_by_location

app/views/results/index.html.erb
```diff
@@ -61,8 +61,7 @@
         <% @courses.each do |course| %>
           <%= render Courses::SearchResultComponent.new(
             course: course,
-            has_sites: @results_view.has_sites?(course),
-            filtered_by_location: @results_view.location_filter?,
+            results_view: @results_view,
```
TobyRet, line 64: I think it is fine to pass in the @results_view object - we also do this on line 17

app/views/results/_sites.html.erb
```diff
@@ -0,0 +1,13 @@
+<% number_of_sites = results_view.sites_count(course) %>
```
TobyRet, line 1: Potentially we could also use a component here also (rather than a partial) 🤔

TobyRet: Not a blocker though ...

TobyRet (changes requested):
Made some small suggestions - almost there though I think

frankieroberto (changes requested):
Can we have a think through of the best content to display where there are multiple school locations listed? I’m not sure "Number of locations: 2" is ideal, as it doesn't help give any sense of the area(s) covered, and also may be inaccurate (if they've not added all the schools, but just a few of them)...

PR #1077, opened 2021-11-03
Don't render the location row on a non location search

JR-G:
### Context

A [support ticket](https://trello.com/c/pZatjybt/859-find-page-not-showing-location) was raised by a provider on how we render the location information on Find, when a candidate searches with 'Across England'.

We're currently rendering the providers address, rather than the site address. As an interim fix we will no longer render the location row, unless the candidate does a locations search.

If a candidate searches 'Across England' or 'By provider' then don't render the locations row.

### Guidance to review

This is only an interim fix until we can come up with a reliable solution.

### Trello card

https://trello.com/c/pZatjybt/859-find-page-not-showing-location

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history


frankieroberto (approved):
👍

PR #1084, opened 2021-11-08
Find feature flags: Step 1 - set up basic auth

TobyRet:
### Context

Currently feature flags on Find are controlled via environment variables. This means we need to redeploy whenever we want to toggle a flag.

Apply on the other hand has the means to toggle flags live in production with the need to redeploy.

We want to be able to store Feature Flag state in Redis and have the means to toggle feature flags via a basic-auth-protected UI

Set up basic auth and create a dummy Feature flags UI page that is protected by basic auth in production

### Guidance to review

The basic auth is enabled in production. To see it in development add the following to your `development.local.yaml` and visit `/feature-flags`

```
basic_auth_enabled: true
basic_auth_username: foo
basic_auth_password: bar
```

### Trello card

https://trello.com/c/raFhJfIq/4123-find-feature-flags-step-1-set-up-basic-auth

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/http_auth_concern.rb
+++ b/app/controllers/concerns/http_auth_concern.rb
@@ -1,0 +1,15 @@
+module HttpAuthConcern
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :http_authenticate
+  end
+
+  def http_authenticate
+    return true unless Settings.basic_auth_enabled
+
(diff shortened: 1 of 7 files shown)


stevehook (approved):
👍 Tested locally and all good. I can't really comment on how 'secure' basic auth is but the implementation here looks fine.

PR #1086, opened 2021-11-12
Create Feature Flags UI and store state in Redis 

TobyRet:
### Context
Currently feature flags on Find are controlled via environment variables. This means we need to redeploy whenever we want to toggle a flag.

Apply on the other hand has the means to toggle flags live in production with the need to redeploy.

We want to be able to store Feature Flag state in Redis and have the means to toggle feature flags via a basic-auth-protected UI

### Changes proposed in this pull request
- Store feature flag state in Redis.
- Allow ability to toggle feature flags from the new Feature Flag UI

feature_flags_final

### Guidance to review
This feature largely mirrors prior art in [Apply](https://github.com/DFE-Digital/apply-for-teacher-training)

- To test it out visit '/feature-flags' and you will be able to activate / deactivate a dummy feature flag (note this is not being used anywhere)
- The next PR will migrate across existing feature flags and remove any old code / config associated with the legacy feature flag logic. 

### Trello card
https://trello.com/c/szHOoTI7/4124-find-feature-flags-step-2-store-feature-flag-state-in-redis

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Tested by running locally

PR #1090, opened 2021-11-15
Migrate feature flags

TobyRet:
### Context

Currently feature flags on Find are controlled via environment variables. This means we need to redeploy whenever we want to toggle a flag.

Apply on the other hand has the means to toggle flags live in production with the need to redeploy.

We want to be able to store Feature Flag state in Redis and have the means to toggle feature flags via a basic-auth-protected UI

- Migrate any current feature flags over to the new system
- Remove any old feature flag code / config

### Guidance to review
- Recommend reviewing per commit 
- Note that before merging, https://github.com/DFE-Digital/find-teacher-training/pull/1086 should be merged first and feature flag state should be set up correctly on on each environment

### Checklist

- [x] Rebased master
- [ ] Cleaned commit history


davidgisbey:
if you toggle maintenance mode on you can't get back to the feature flag page. I think we need to add an exemption a skip before action to the controller

```
  skip_before_action :redirect_to_maintenance_page_if_flag_is_active

```

davidgisbey (approved):

TobyRet:
> if you toggle maintenance mode on you can't get back to the feature flag page. I think we need to add an exemption a skip before action to the controller
> 
> ```
>   skip_before_action :redirect_to_maintenance_page_if_flag_is_active
> ```
Good spot @davidgisbey 

Have made the change - https://github.com/DFE-Digital/find-teacher-training/pull/1090/commits/293550109b58867c955843488da4e41658f3f62a

stevehook (commented):
I had a look at the errors in `rails assets:precompile`. We don't need to initialise BigQuery for asset precompilation so I guess there is no need to run the initialiser though I can't see an easy way to 'switch it off'. There used to be an option in Rails like `config.assets.initialize_on_precompile = false` but that's been dropped in recent versions.

Is there are way to programmatically find out whether an initialiser is running for asset precompilation. If there is we could use that to just return immediately from `initializers/big_query.rb`. Failing that we could handle the exception in the same file, but that would mean that we could miss genuine Redis connectivity issues in future.

TobyRet:
> I had a look at the errors in `rails assets:precompile`. We don't need to initialise BigQuery for asset precompilation so I guess there is no need to run the initialiser though I can't see an easy way to 'switch it off'. There used to be an option in Rails like `config.assets.initialize_on_precompile = false` but that's been dropped in recent versions.
> 
> Is there are way to programmatically find out whether an initialiser is running for asset precompilation. If there is we could use that to just return immediately from `initializers/big_query.rb`. Failing that we could handle the exception in the same file, but that would mean that we could miss genuine Redis connectivity issues in future.

@stevehook - looking at Apply it appears the big query initializer on checks for the existence of `BIG_QUERY_API_JSON_KEY` env var - the https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/config/initializers/bigquery.rb#L5

We could do something similar but then could get into a situation where the env var is not present but the big query feature flag is turned on ... But I'm assuming the app will raise an error if it makes and attempt to send events when big query has not been initialised?

stevehook:
> > I had a look at the errors in `rails assets:precompile`. We don't need to initialise BigQuery for asset precompilation so I guess there is no need to run the initialiser though I can't see an easy way to 'switch it off'. There used to be an option in Rails like `config.assets.initialize_on_precompile = false` but that's been dropped in recent versions.
> > Is there are way to programmatically find out whether an initialiser is running for asset precompilation. If there is we could use that to just return immediately from `initializers/big_query.rb`. Failing that we could handle the exception in the same file, but that would mean that we could miss genuine Redis connectivity issues in future.
> 
> @stevehook - Thanks for checking this out. Looking at Apply it appears the big query initializer checks for the existence of `BIG_QUERY_API_JSON_KEY` env var - https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/config/initializers/bigquery.rb#L5
> 
> We could do something similar but then could get into a situation where the env var is not present but the big query feature flag is turned on ... But I'm assuming the app will raise an error if it makes and attempt to send events when big query has not been initialised?

If `BIG_QUERY_API_JSON_KEY` is missing when running the asset pipeline this is a better idea than anything I suggested.

PR #1094, opened 2021-11-17
Componentise the results page

stevehook:
### Context

The views in Find are complex with lots of logic that is tested via feature specs or not at all. The intention of this PR (and others related) is to componentise those views using view components, to encapsulate the logic that is currently in view templates and to add tests alongside those components to test that logic in isolation.

- [x] Add a `componentised_results` feature flag. (This will need to be moved to the new feature flag implementation).
- [x] New `Results` namespace for the new components
- [x] Add a new `index.html.erb` template and rename the old one so that it can be easily dropped when the above feature flag is retired. The new template will delegate to a new `Results::ResultsComponent`.
- [x] Implement sub-components of `Results::ResultsComponent`
- [x] Integrate with new feature flag implementation

These will be handled in follow-up PR(s):
- Identify and implement subcomponents of `Results::FilterComponent` (Currently this consists of several partials).
- Break out logic from `ResultsView` - should some of it be in `ResultsComponent`?
- Revisit feature specs - are they all still necessary now?

### Guidance to review

I'm keeping this as draft for now because I think it needs to be released after the feature flag re-implemenatation. It's ready for review though.

### Trello card

https://trello.com/c/6I7jkwbg/4127-componentise-the-results-page

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


app/components/results/results_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+</h1>
+
+<div class="app-filter-toggle"></div>
+
+<div class="app-filter-layout">
+  <div class="app-filter-layout__filter">
+    <%= render Results::FilterComponent.new(results: results) %>
+  </div>
+  <div class="app-filter-layout__content">
+    <% if results.no_results_found? %>
```
TobyRet, line 14: Should the responsibility for deciding whether to render or not be moved to the component's `#render` method?

TobyRet (commented):
This looks great. 
Now that we have unit test coverage does this potentially mean that feature specs can be trimmed?

TobyRet:
> This looks great. Now that we have unit test coverage does this potentially mean that feature specs can be trimmed?

Ah just seen you raised this as a question ...

Could be easier to prune the feature specs as we go, but not a blocker

TobyRet (approved):

PR #1101, opened 2021-11-29
[4128] - Componentise course page

TobyRet:
### Context

When Find was originally ported from C# to Ruby we did not use the most execellent 'view components' gem as Apply has done. As a result the views in Find are complex with lots of logic that is tested via feature specs or not at all! 

Introducing view components will allow us to more easily test view logic via unit specs rather than feature specs 

- extract view components and add unit tests
- remove any duplicated legacy logic (specs, partials etc)

### Guidance to review
- Recommend reviewing per commit
- A lot if the partials were checking that an attribute was `present?` before rendering - I have not added test coverage (via component specs) in these instances as it feels unnecessary - what do people think?
- There are still some simple partials remaining in the course show page - I did not extract these as I felt there was no benefit. Is it ok to have partials and components living side by side?

### Trello card
https://trello.com/c/U2zWy1DD/4128-componentise-the-course-page

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


app/components/courses/about_schools_component.html.erb
```diff
@@ -0,0 +1,53 @@
+<% if course.how_school_placements_work.present? ||
+  course.program_type == 'higher_education_programme' ||
+  course.program_type == 'scitt_programme' ||
+  course.site_statuses.map(&:site).uniq.many? %>
```
avinhurry, line 4: There's a bit of duplication here with the `render?` method below in `app/components/courses/about_schools_component.rb`. 

Can we reuse that logic?

avinhurry (commented):
Nice work, this looks a lot neater. Just one small comment in line. I think it would nice to refactor the remaining partials for consistency as tech debt further down the line.

Also, do we need to make these changes to preview on Publish?

app/components/courses/about_schools_component.html.erb
TobyRet: Good spot - the duplicated code should not be in the erb file - have removed

TobyRet:
> Nice work, this looks a lot neater. Just one small comment in line. I think it would be nice to refactor the remaining partials for consistency as tech debt further down the line.
> 
> Also, do we need to make these changes to preview on Publish?

Yes there are some partials remaining but this is intentional. I've extracted components from partials when they contained logic or behaviour or some sort of testable complexity. The partials that remain are 'dump' and simply display html - there would be little benefit to converting them to components IMO.

There has been no change to the course the preview that would result it in rendering differently from the one in Publish however @malcolmbaig and his team may still be interested this PR all the same.

app/components/courses/apply_component.html.erb
```diff
@@ -0,0 +1,36 @@
+<div class="govuk-!-margin-bottom-8" id="section-apply">
+  <% if CycleTimetable.mid_cycle? %>
+    <% if has_vacancies? == 'Yes' %>
```
stevehook, line 3: Having a `has_xxx?` method that returns a string is the kind of thing that would trip me up! I'm a bit confused because in the tests it looks like it's stubbed with `true` or `false`.

stevehook (approved):
This looks really great.

I had one little question inline.

app/components/courses/apply_component.html.erb
TobyRet: Yes good point @stevehook - we shouldn't be decorating the `#has_vacancies?` 
I've made the change - https://github.com/DFE-Digital/find-teacher-training/pull/1101/commits/cac537401e6eca5dd24fcb5ae3931d323bb2a243

PR #1118, opened 2021-12-22
[4209] - Wire up provider search to the new flow

TobyRet:
### Context 

We have wired up searching by 'location' and 'across England' to the new search flow. We now need to do the same for searching by 'provider'. 

### Changes in the pull request 
- Wire up provider search to the new flow
- Add fix so the the provider search value is rehydrated on the start page when clicking back 

### Guidance for pull request 
- Visit '/feature-flags' and activate the 'new_search_flow' feature flag 

### Trello 
https://trello.com/c/coZ9yN3R/4209-primary-secondary-fe-flow-apply-to-provider-search-journey

PR #1129, opened 2021-12-30
[4234] - WIP: Refactor start page

TobyRet:
### Context

IMPORTANT - https://github.com/DFE-Digital/find-teacher-training/pull/1118 needs to be merged first

We want to bring the start page ('/') in line with the new search flow journey.

We want to remove the use of the 'flash object' for validation errors on this page and rely on the form object only. 

We also can remove the `flash[:start_wizard]` references as candidates should be directed down through the new search flow (selecting age group and subjects) whenever they edit their search query rather than redirected straight to the results page. 

Note this is still a work in progress.

Things done so far:
- When the `new_search_flow` feature flag is activated the user is directed to a new view / controller 
- A new form has been created, validation is working
- The location search part of the journey has been wired up

Things to do:
- wire up the 'across England' and 'provider' search journeys
- check code coverage and add specs if req'd

Other things we may want to do:
- Currently we are still using the following legacy query params `l` (search type), `lq` (location query) and `query` (provider query). We should probably rename these to `search-type`, `location-query` and `search-query` however it is probably worth waiting until the `new_search_flow` feature flag is removed

### Trello card
https://trello.com/c/5GelL6MG/4234-draft-find-refactor-start-page

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/result_filters/location_controller.rb
+++ b/app/controllers/result_filters/location_controller.rb
@@ -8,6 +8,11 @@ module ResultFilters
     def new; end
 
     def start
+      if FeatureFlag.active?(:new_search_flow)
+        redirect_to start_path
+        return
+      end
+
       flash[:start_wizard] = true
     end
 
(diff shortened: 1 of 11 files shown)


avinhurry:
Closing until we're ready to make more changes to find

PR #1143, opened 2022-01-07
Copy smoke tests from external repo

frankieroberto:
This just copies across the "smoke tests" (written using [cypress](https://www.cypress.io)) from the external [find-teacher-training-tests](https://github.com/DFE-Digital/find-teacher-training-tests) repository.

This should enable us to keep the tests more in sync with the code, for example updating both in a single pull request.


.github/workflows/smoke-test.yml
```diff
@@ -21,10 +21,6 @@ jobs:
 
       - name: Checkout
```
duncanjbrown, line 22: Could we lose this bit too?

.github/workflows/smoke-test.yml
```diff
@@ -54,7 +50,7 @@ jobs:
           if-no-files-found: ignore
           retention-days: 7
 
-      - name: 'Nofiy #twd_apply_tech on failure'
+      - name: 'Notify #twd_apply_tech on failure'
```
duncanjbrown, line 53: ⭐

.github/workflows/smoke-test.yml
frankieroberto: I was assuming it'd still need to checkout the repo, in order to get the actual tests?

duncanjbrown: ah i see — sorry, thought this was the action for the main bit, ignore me!

frankieroberto:
@duncanjbrown let's merge this, and do a follow-up PR which updates the tests and turns on the feature flag by default?

duncanjbrown (approved):

PR #1149, opened 2022-01-11
[487] Increase deploy workflow timeout

saliceti:
### Context
The timeout was set to 15 minutes at the job level but this is often reached when multiple PRs are merged quickly.

We now use the same configuration as in the other services: set it to 30 minutes at the wait step level and only lock the workflow runs when on the same branch.

### Guidance to review
Same configuration as Register, etc

### Trello card
https://trello.com/c/GS44pWv6

### Checklist

- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,7 +30,6 @@ jobs:
   prepare-matrix:
     name: Prepare Environment Matrix
     runs-on: ubuntu-latest
-    timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


chelnak (approved):

saliceti:
[Find deploy workflow timeout](https://trello.com/c/GS44pWv6/487-find-deploy-workflow-timeout)

PR #1213, opened 2022-02-11
confusing sentence that was removed elsewhere

EmmaFrith:
we removed this sentence elsewhere because it's confusing. but we forgot this instance of it.


frankieroberto (approved):
Doh. How did we miss this? 🤦

PR #1231, opened 2022-02-24
Use email address as link text for mailto

frankieroberto:
This makes it clearer that the link will open an email client, which might otherwise be unexpected and annoying, for example if the link opens Outlook but you use Gmail.

Using the email address as link text makes it easier for users to copy and paste the email address into their preferred email app if needed.

### Before

Screenshot 2022-02-24 at 13 45 29

Screenshot 2022-02-24 at 13 49 14

### After

Screenshot 2022-02-24 at 13 46 02

Screenshot 2022-02-24 at 13 49 44

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -100,7 +100,7 @@
           colour: environment_colour,
         },
       ) do %>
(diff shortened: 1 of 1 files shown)


tomas-stefano (approved):
🚀

MylesJarvis (approved):

PR #1245, opened 2022-03-01
simplify fees and financial support section

EmmaFrith:
this is because:

- policies may change yearly and so keeping content up to date is difficult 
- scholarship bodies set their own eligibility criteria which we have not verified 
- all the content is on GIT, and may be changed frequently on GIT, so we don't want to risk being inconsistent or inaccurate

### DIFF
--- a/app/components/courses/financial_support/scholarship_and_bursary_component.html.erb
+++ b/app/components/courses/financial_support/scholarship_and_bursary_component.html.erb
@@ -16,15 +16,7 @@
     <% end %>
 
     <p class="govuk-body">
-      To qualify for a scholarship you’ll need a degree of 2:1 or above in <%= subject_name %> or a related subject. For a bursary you’ll need a 2:2 or above in any subject.
-    </p>
-
-    <p class="govuk-body">
-      You cannot claim both a bursary and a scholarship - you can only claim one.
-    </p>
-
-    <p class="govuk-body">
-      <%= govuk_link_to 'Find out more about bursaries and scholarships', t('get_into_teaching.url_bursaries_and_scholarships') %>. You do not need to apply for a bursary - if you’re eligible, you’ll automatically start receiving it once you begin your course.
+      <%= govuk_link_to 'Find out about bursaries or scholarships, including eligibilty criteria', t('get_into_teaching.url_bursaries_and_scholarships') %>.
     </p>
 
     <p class="govuk-body">


app/components/courses/financial_support/scholarship_and_bursary_component.html.erb
```diff
@@ -16,15 +16,7 @@
...
-      To qualify for a scholarship you’ll need a degree of 2:1 or above in <%= subject_name %> or a related subject. For a bursary you’ll need a 2:2 or above in any subject.
-    </p>
-
-    <p class="govuk-body">
-      You cannot claim both a bursary and a scholarship - you can only claim one.
-    </p>
-
-    <p class="govuk-body">
-      <%= govuk_link_to 'Find out more about bursaries and scholarships', t('get_into_teaching.url_bursaries_and_scholarships') %>. You do not need to apply for a bursary - if you’re eligible, you’ll automatically start receiving it once you begin your course.
+      <%= govuk_link_to 'Find out about bursaries or scholarships, including eligibilty criteria', t('get_into_teaching.url_bursaries_and_scholarships') %>.
```
EmmaFrith, line 19:
```suggestion
      <%= govuk_link_to 'Find out about bursaries or scholarships, including eligibility criteria', t('get_into_teaching.url_bursaries_and_scholarships') %>.
```

app/views/courses/financial_support/_scholarship_and_bursary.html.erb
```diff
@@ -16,15 +16,7 @@
...
-      To qualify for a scholarship you’ll need a degree of 2:1 or above in <%= course.subject_name %> or a related subject. For a bursary you’ll need a 2:2 or above in any subject.
-    </p>
-
-    <p class="govuk-body">
-      You cannot claim both a bursary and a scholarship - you can only claim one.
-    </p>
-
-    <p class="govuk-body">
-      <%= govuk_link_to 'Find out more about bursaries and scholarships', t('get_into_teaching.url_bursaries_and_scholarships') %>. You do not need to apply for a bursary - if you’re eligible, you’ll automatically start receiving it once you begin your course.
+      <%= govuk_link_to 'Find out about bursaries and scholarships, including eligibility criteria', t('get_into_teaching.url_bursaries_and_scholarships') %>. You do not need to apply for a bursary - if you’re eligible, you’ll automatically start receiving it once you begin your course.
```
EmmaFrith, line 19:
```suggestion
      <%= govuk_link_to 'Find out about bursaries and scholarships, including eligibility criteria', t('get_into_teaching.url_bursaries_and_scholarships') %>.
```

frankieroberto:
@EmmaFrith interesting! I agree this section could be simplified.

I’m not sure about linking to GIT though. The page being linked to doesn't actually give you much more information, other than:

* explaining what scholarships and bursaries are ("tax-free amounts of money you receive to train in certain subjects. You do not need to pay them back.") – perhaps we should do this too, if we think not everyone will know what they are?
* giving the general degree eligibility for bursaries only ("a bachelor’s degree, class 2:2 (honours) or higher, a PhD or a master’s degree".

The rest is then links to other sources.

I wonder whether we could skip GIT and link direct to those sources? We could also then be a bit more specific, and only link to the relevant scholarship page (eg Royal Society of Chemistry) rather than listing all 4?  The downside is that the page for [bursary eligibility](https://www.gov.uk/government/publications/funding-initial-teacher-training-itt/funding-initial-teacher-training-itt-academic-year-2022-to-2023) is not super candidate-friendly, and also doesn't even seem to be 100% complete:

> Further guidance on postgraduate bursaries and scholarships will be available in the [Initial teacher training (ITT) bursary: funding manual](https://www.gov.uk/government/publications/initial-teacher-training-itt-bursary-funding-manual) in due course.

I've also just realised that there’s guidance that we should [avoid anchor links](https://www.gov.uk/guidance/content-design/links#avoid-anchor-links) as they can be disorientating – which I suspect is doubly the case when we’re linking to GIT as that's a whole separate brand?

Perhaps all this indicates that there should be a new dedicated page for bursaries somewhere (on GOV.UK?) which is candidate facing and gives the essential information of what the eligibility criteria is, how to apply, and so on? 🤔

EmmaFrith:
Hey @frankieroberto, so GIT is actually improving their content atm!

PR #1246, opened 2022-03-02
Refine conditions for smoke test execution

NickGraham101:
### Context

The smoke test results are uploaded even if a previous step fails, this is desirable if that step is the smoke test execution but if  an earlier step fails the reported results are confusing.  If the smoke test fails but the upload succeeds the `job.status` is successful, as this context property is used to report on whether the deployment has succeeded deployments with failed tests are reported as successful.

This change ensures that the results are uploaded only if the tests are executed and deployment outcomes are correctly reported:
- the `outcome` property of the `run_cypress` step is used as the condition for the upload step
- a step is added to evaluate the outcome of all preceding steps
- all steps prior to this are given an `id` to ensure that their `outcome` property is accessible

https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

### Guidance to review

A failing step was temporarily added as part of development to test the process.  The failing step can be seen [here](https://github.com/DFE-Digital/find-teacher-training/runs/5406471806?check_suite_focus=true#step:3:319) and the deployment result is logged [here](https://github.com/DFE-Digital/find-teacher-training/runs/5406471806?check_suite_focus=true#step:3:319).  This is displayed in the [Environment page](https://github.com/DFE-Digital/find-teacher-training/deployments) of the GitHub UI like this:

image

- Consider whether this is the desired behaviour

### Trello card

https://trello.com/c/grpMynpF

### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -23,6 +23,7 @@ inputs:
 runs:
   using: composite
   steps:
+    # all steps prior to the "Evaluate job status" step require an id to ensure that we correctly report the deployment outcome
     - name: Start ${{ inputs.environment }} Deployment
       uses: bobheadxi/deployments@v1
       id: deployment
@@ -33,6 +34,7 @@ runs:
         ref: ${{ inputs.sha }}
 
     - name: Set Environment variable
+      id: set_env_var
       shell: bash
       run: |
         if [ -n "${{ inputs.pr }}" ]; then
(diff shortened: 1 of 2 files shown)


.github/actions/deploy/action.yml
```diff
@@ -33,6 +33,7 @@ runs:
         ref: ${{ inputs.sha }}
 
     - name: Set Environment variable
+      id: set_env_var
```
saliceti, line 36: It becomes critical to add an id to each step, caould we add a generic comment to this workflow to emphasise this so any future step are created with an id

.github/actions/deploy/action.yml
```diff
@@ -106,14 +113,24 @@ runs:
         if-no-files-found: ignore
         retention-days: 7
 
+    - name: Evaluate job status
+      if: always()
+      shell: bash
+      run: |
+        if [[ ${{ join(steps.*.outcome, ',') }} =~ "failure" ]]; then
+          echo JOB_STATUS="failure" >> $GITHUB_ENV;
```
saliceti, line 121: I think semi colons are not necessary if there is a line break

saliceti (approved):

.github/actions/deploy/action.yml
NickGraham101: removed

PR #1264, opened 2022-03-16
Stop deployment status marking deployment inactive

NickGraham101:
### Context

The deployment status update task by default (in version 1.1.0) actively marks previous deployments as inactive, this results in a large number of API calls which in turn causes a rate limiting failure.

This change disables that function and explicitly enables the builtin GitHub behaviour that marks all previously successful deployments as inactive.

### Checklist

- [ ] Rebased `main`
- [ ] Cleaned commit history


saliceti (approved):

saliceti:
[Github rate limiting](https://trello.com/c/hLfLCJ9I/540-github-rate-limiting)

PR #1290, opened 2022-04-12
move course button and no vacancies warning text to top of course pages

EmmaFrith:
Currently the apply button, and no vacancies warning, appear right at the bottom of the page. 

Users might struggle to fin them because our pages are so long. 

Move these to the top of the page, as well as the bottom, so they're more visible.

Screenshot 2022-04-12 at 15 07 34

Screenshot 2022-04-12 at 15 08 01

## Question

I couldn't see how this change looks at the end of the cycle, when other content replaces the apply button. See screenshot for the content that replaces the button (also described in the design history entry here: https://bat-design-history.netlify.app/find-teacher-training/changes-to-find-at-the-end-of-2021-cycle/#the-deadline-to-apply-again-has-passed). 

How do I simulate the end of cycle locally, so I can see this in context?

Screenshot 2022-04-12 at 15 11 07

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -24,6 +24,8 @@
 
   <%= render Courses::SummaryComponent.new(course) %>
 
+    <%= render Courses::ApplyComponent.new(course) %>
+
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= render Courses::ContentsComponent.new(course) %>
@@ -60,7 +62,6 @@
 
       <%= render partial: 'courses/advice' %>
 
-      <%= render Courses::ApplyComponent.new(course) %>
     </div>
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


app/views/courses/show.html.erb
```diff
@@ -61,6 +63,7 @@
       <%= render partial: 'courses/advice' %>
 
       <%= render Courses::ApplyComponent.new(course) %>
```
avinhurry, line 65: Looks like you forgot to remove the component at the bottom, so I think there are now two on the page. This should also fix your failing test 😄

avinhurry (changes requested):
Nice work. I think this change makes things a lot better! One comment in line which will also fix your failing test.

I've added a screenshot below of what it looks like if Apply is closed. Is that what you were after?

image

avinhurry:
Superceded with #1292

PR #1292, opened 2022-04-13
Move warning text to the top

avinhurry:
### Context

Moving the warning text which appears if a course does not have vacancies to help make this more clear, earlier on to candidates but keeping the Apply button at the bottom for now.

- Move warning text to the top
- Keep the apply button and end of cycle text at the bottom

### Guidance to review

Is the logic all correct?

### Trello card

https://trello.com/c/XIgsStek/4570-design-move-course-full-warning-to-top-of-find-pages

### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -24,6 +24,8 @@
 
   <%= render Courses::SummaryComponent.new(course) %>
 
+  <%= render Courses::ApplyComponent.new(course) unless course.has_vacancies? %>
+
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= render Courses::ContentsComponent.new(course) %>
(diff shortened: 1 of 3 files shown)


EmmaFrith (approved):
yay! thanks so much @avinhurry

app/views/courses/show.html.erb
```diff
@@ -60,7 +62,7 @@
 
       <%= render partial: 'courses/advice' %>
 
-      <%= render Courses::ApplyComponent.new(course) %>
+      <%= render Courses::ApplyComponent.new(course) if course.has_vacancies? || !CycleTimetable.mid_cycle? %>
```
stevehook, line 65: Does this mean that you'll have two elements with `id=section-apply` when `course.has_vacancies?` is false and `CycleTimetable.mid_cycle?` is false?

stevehook (approved):
Logic seems sound to me. I just had one question inline.

app/views/courses/show.html.erb
avinhurry: The two elements should never render at the same time I think because of the additional logic in `apply_component.html.erb` so there should only ever be one element with `id=section-apply`

PR #1305, opened 2022-04-22
Use dfe-analytics

duncanjbrown:
### Context

Replace this app's custom BigQuery integration with the `dfe-analytics` gem


duncanjbrown:
Deploying this to a review app so we can test it works 😅

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apk add --update --no-cache tzdata && \
     echo "Europe/London" > /etc/timezone
 
 RUN apk add --update --no-cache --virtual runtime-dependances \
-    yarn shared-mime-info
+    yarn shared-mime-info git
```
misaka, line 10: What is this needed in the Docker container for? I can't see where git is used there ...

duncanjbrown: Because we're pulling from GitHub until we get RubyGems

misaka: Ah, for when we run `bundle` in the Docker container. Of course. 👍

misaka (approved):

PR #1313, opened 2022-04-28
Redesign promoted event link

frankieroberto:
Move this above the "Sorted by" component, and redesign to use a background colour instead of inset text.

This promoted event link was added as an experiment, to see if it made an impact.

The inset text component is known to sometimes be ignored by users, and should probably be reserved for introducing alternative "voices" (for example quoting a message from another organisation).

Moving the link up also brings the "Sorted by" component back closer to the content it describes.

🗂️  [Trello card](https://trello.com/c/HgsSuqHr/2971-update-find-banner-for-event-promotion)

### Before

Screenshot 2022-04-28 at 15 37 12

### After

Screenshot 2022-04-28 at 15 36 53

### DIFF
--- a/app/components/results/results_component.html.erb
+++ b/app/components/results/results_component.html.erb
@@ -11,14 +11,14 @@
     <%= render Results::FilterComponent.new(results: results) %>
   </div>
   <div class="app-filter-layout__content">
-    <%= render Results::NoResultsComponent.new(results: results) %>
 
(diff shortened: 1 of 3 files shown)


MylesJarvis (approved):
LGTM - I'll merge this after I dig through the data with Gemma tomorrow.

EmmaFrith (approved):

PR #1315, opened 2022-04-29
Bugfix: always go to the age group page after first page

frankieroberto:
This fixes a bug where if you return to the first page after using the back button from the results page, then change your location and press continue, you end up on the results page with ALL courses selected.

Expected behaviour is to be taken back through the age group and subject flow again.

🗂️ [Trello card](https://trello.com/c/61oDmqTZ/4595-find-search-flow-doesnt-like-back-button-followed-by-an-alternative-search)

## Screenshot showing bug

Screenshot 2022-04-29 at 13 20 52


avinhurry (approved):
Hmm, I was unable to replicate this bug locally 🤔, although I could in production. Seems fine on the review app and code looks good 👍🏼

PR #1316, opened 2022-04-29
Remove legacy subject search page

frankieroberto:
This page is no longer used within the flow, but is still available if you visit the URL `/start/subject`. We should remove it and redirect to homepage.

🗂️ [Trello card](https://trello.com/c/ezFc6AjK/4384-if-you-navigate-directly-to-a-page-in-the-new-search-flow-on-find-it-renders-the-old-accordion-design)


tomas-stefano (approved):
LGTM

PR #1323, opened 2022-05-06
Add default username and password in development mode

frankieroberto:
This was missing for some reason, which prevented access to the feature flags interface.

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -5,3 +5,5 @@ valid_referers:
 log_level: debug
 redis_url: redis://localhost:6379/0
 background_jobs: # scheduled jobs run on all environments except dev and test
+basic_auth_username: replace_me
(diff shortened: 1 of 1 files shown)


duncanjbrown:
would we do better to disable auth in development mode? it doesn't serve any purpose

frankieroberto:
> would we do better to disable auth in development mode? it doesn't serve any purpose

Probably? I also wondered about moving this to env variables instead of being hardcoded with the config files that are part of Git?

We might also want to consider moving away from basic authentication altogether, in favour of an HTML-based password screen, as it'd more usable/accessible, and some organisations / browsers are starting to drop support for basic auth: https://github.com/alphagov/govuk-prototype-kit/issues/1019 ?

defong (approved):
LGTM

PR #1333, opened 2022-05-13
Tweak service name

frankieroberto:
Renames the service from 'Find postgraduate teacher training' to 'Find teacher training courses'.

Rationale:

* Apply dropped the word 'postgraduate' long ago, after finding that it wasn't necessary, but we didn't go back and change Find
* It's good to be consistent between Find and Apply
* Find contains Further education courses, which don't always require a degree and aren't postgraduate
* "postgraduate" is an awkward word which not everyone may recognise, particularly non-UK applicants

### DIFF
--- a/app/job/slack_notification_job.rb
+++ b/app/job/slack_notification_job.rb
@@ -25,7 +25,7 @@ private
     end
 
     payload = {
-      username: 'Find postgraduate teacher training',
+      username: 'Find teacher training courses',
       channel: slack_channel,
       text: slack_message,
       mrkdwn: true,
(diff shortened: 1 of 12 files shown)


defong:
@frankieroberto is this :headstone: ?

frankieroberto:
@defong that's up to your team! I think it's definitely worth considering tweaking the service name in this way, for the reasons given in the PR above.

defong:
> I'd close this for now @defong - it needs a wider team conversation before doing something like this


Closing this then

PR #1423, opened 2022-07-28
346 Content Security Policy

goodviber:
### Context
The Git Into Teaching marketing team would like to setup google ads to point to the Find website instead of the current Get into teaching website to improve conversation tracking.

We should setup a content security policy on Find to make sure it only executes script we expect.

What needs to be done
Add a CSP for Find which whitelists scripts from:

Google analytics
Google tag manager
Success
We have a CSP setup

CSP outline, moved one inline JS script to partial
### Guidance to review
Check we are not seeing any content-security-policy related console errors in the deploy app
### Trello card
[Trello](https://trello.com/c/hWQjYsWh/346-setup-csp-for-external-scripts)
### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history


config/initializers/content_security_policy.rb
```diff
@@ -4,23 +4,37 @@
...
+    policy.font_src    :self, :data
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self,
+                       :unsafe_inline, # Backwards compatibility; ignored by modern browsers as we set a nonce for scripts
+                       'https://www.google-analytics.com',
+                       'https://www.googletagmanager.com'
+
+    policy.connect_src :self,
+                       'https://api.postcodes.io', # Not sure we need this?
```
darokel, line 19: I don't think we need this or the one below `https://stats.g.doubleclick.net'`

darokel (approved):
Let's get those two removed and this should be good to go. Would be good to double check in prod that the GA traffic is still coming through after this gets merged

defong:
[Setup CSP for external scripts](https://trello.com/c/hWQjYsWh/346-setup-csp-for-external-scripts)

PR #1435, opened 2022-08-04
[365] Update the Ruby version to 3.1.0

darokel:
### Context

https://trello.com/c/uzlYuEsR/365-update-find-to-ruby-3

Updates the Ruby version to be consistent with Publish

### Guidance to review

The app still works as expected

### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history


.rubocop.yml
```diff
@@ -11,6 +11,7 @@ Govuk/GovukSubmit:
   Enabled: false
 
 AllCops:
+  TargetRubyVersion: 3.1.0
```
avinhurry, line 14: What's this for? Do we need it, doesn't rubocop get the ruby version from `gemfile.lock`?

avinhurry (commented):
Looks good, and works well! I just had one question RE rubocop inline

.rubocop.yml
darokel: You'd think that but the CI has other ideas. I'd prefer to be explicit so it doesn't complain

Screenshot 2022-08-08 at 09 46 59

avinhurry (approved):

PR #1440, opened 2022-08-08
346 csp external scripts v2

goodviber:
### Context
The Git Into Teaching marketing team would like to setup google ads to point to the Find website instead of the current Get into teaching website to improve conversation tracking.

We should setup a content security policy on Find to make sure it only executes script we expect.

What needs to be done
Add a CSP for Find which whitelists scripts from:

Google analytics
Google tag manager
Success
We have a CSP setup

Tech notes
The teaching vacancies app is a good source for an example as they've had to accommodate a similar need for marketing [teaching-vacancies/content_security_policy.rb at main · DFE-Digital/teaching-vacancies](https://github.com/DFE-Digital/teaching-vacancies/blob/main/config/initializers/content_security_policy.rb)

Add CSP

### Guidance to review
NB self does not include subdomains, eg. 'https://assets.find-postgraduate-teacher-training.service.gov.uk' so they must be wildcarded

### Trello card
[Trello](https://trello.com/c/hWQjYsWh/346-setup-csp-for-external-scripts)
### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history

### DIFF
--- a/app/views/layouts/_add_js_enabled_class_to_body.html.erb
+++ b/app/views/layouts/_add_js_enabled_class_to_body.html.erb
@@ -1,0 +1,3 @@
+<%= javascript_tag(nonce: true) do -%>
+  document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
+<% end -%>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,9 +20,7 @@
   </head>
 
   <body class="govuk-template__body ">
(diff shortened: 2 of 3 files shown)


SamJamCul:
I'm getting this warning error thing when loading up the review app - something to be concerned about?
image

goodviber:
> I'm getting this warning error thing when loading up the review app - something to be concerned about? image

I think this is OK, it is describing the expected behaviour...I think

SamJamCul (approved):

PR #1651, opened 2022-12-16
[870] Update course search results title

darokel:
### Context

We recently did some accessibility work to [fix the page titles across Find](https://trello.com/c/GaFrnjBD). As part of this work, we discovered that the course results list page title could be improved.

- Extract logic for building out the results page title into a component

### Guidance to review

Check the title on the results page builds up the correct string for:

- Location search
- England search
- Provider search

### Trello card

https://trello.com/c/oAdyZoHU/870-update-courses-results-list-title

### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history

### DIFF
--- a/app/components/result_summary_component.rb
+++ b/app/components/result_summary_component.rb
@@ -1,0 +1,55 @@
+class ResultSummaryComponent < ViewComponent::Base
+  include ViewHelper
+
+  attr_reader :results
+
+  def initialize(results:)
+    @results = results
+  end
+
+  def call
+    [
+      course_found_text,
+      subject_names,
+      with_location_text,
+      with_england_text,
+      with_provider_text,
+    ].compact.join(' ')
+  end
+
+private
+
+  def subject_names
+    results.subjects.map { |subject| subject[:subject_name].downcase }.to_sentence(last_word_connector: ' and ')
+  end
+
+  def course_found_text
+    case results.course_count
+    when 0
+      'No'
+    when 1
+      '1'
+    else
+      number_with_delimiter(results.course_count)
+    end
+  end
+
+  def course_string_modifier
+    return 'courses found' if results.course_count.zero?
+    return 'course' if results.course_count == 1
+
+    'courses'
+  end
+
+  def with_location_text
+    "#{course_string_modifier} in #{results.location_search}" if results.location_filter?
+  end
+
+  def with_england_text
+    "#{course_string_modifier} in England" if results.england_filter?
+  end
+
+  def with_provider_text
+    "#{course_string_modifier} from #{results.provider}" if results.provider_filter?
+  end
+end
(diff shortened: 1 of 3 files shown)


darokel:
[Update courses results list <title>](https://trello.com/c/oAdyZoHU/870-update-courses-results-list-title)

app/components/result_summary_component.rb
```diff
@@ -0,0 +1,55 @@
+class ResultSummaryComponent < ViewComponent::Base
```
defong, line 1: I'm not sure the merit of using a `ViewComponent` to manipulate text maybe a `helper`?

Maybe the lack of html is a telltale sign?

app/components/result_summary_component.rb
```diff
@@ -0,0 +1,55 @@
...
+      with_location_text,
+      with_england_text,
+      with_provider_text,
+    ].compact.join(' ')
+  end
+
+private
+
+  def subject_names
+    results.subjects.map { |subject| subject[:subject_name].downcase }.to_sentence(last_word_connector: ' and ')
```
defong, line 23: Proper nouns needs correct casing

defong (changes requested):

app/components/result_summary_component.rb
darokel: View components can build up return values via `call`, you don't need a corresponding html file - it's just a ruby class at the end of the day. Also there's a little too much logic to be put into a helper.

app/components/result_summary_component.rb
darokel: There's a separate ticket for this (which Simon knows about). We can cross port the approach implemented and handle this bit too. I'll add it to that ticket

app/components/result_summary_component.rb
defong: Then maybe a simple `service` as it feels over engineered using a view component.

darokel: done

app/services/result_summary_service.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def initialize(results:)
+    @results = results
+  end
+
+  class << self
+    def call(...)
+      new(...).call
+    end
+  end
```
defong, line 14:
```suggestion
  include ServicePattern
```

defong (approved):

simonwhatley (changes requested):
Where the subject contains a noun (for example, Primary with English, English, French, German, etc.), the noun should be in title case.

For example: 
- `123 primary with English courses in England`
- `456 French and German courses in England`
- `789 French, German and Spanish courses in England`
- `321 dance and English courses in England`

darokel:
> Where the subject contains a noun (for example, Primary with English, English, French, German, etc.), the noun should be in title case.
> 
> For example:
> 
> * `123 primary with English courses in England`
> * `456 French and German courses in England`
> * `789 French, German and Spanish courses in England`
> * `321 dance and English courses in England`

We've got a sep ticket to cover that requirement in Publish. We'll cross port the approach for this bit too rather than do things twice/differently.

simonwhatley (approved):

PR #1691, opened 2023-01-23
Remove qa routes from code

RMcVelia:
### Context
For the find migration to publish
Routes will be removed from find tf state and code so they can be managed from publish instead

Steps are 
- take a note of the route guids as they will be used for import in publish
- remove routes from tf state (terraform state rm) 
- remove route mappings from webapp in tf state (terraform pull, edit, terraform push)
- merge this pr

Add variable disable_routes
If variable not null for an env, then routes will not be created

### Guidance to review

make qa deploy-plan

### Trello card

https://trello.com/c/txVxNL4S/744-findpub-merge

### Checklist

- [ ] Rebased `main`
- [ ] Cleaned commit history

### DIFF
--- a/terraform/modules/paas/main.tf
+++ b/terraform/modules/paas/main.tf
@@ -30,7 +30,7 @@ resource cloudfoundry_app web_app {
   dynamic "routes" {
     for_each = local.web_app_routes
     content {
-      route = routes.value
+      route = routes.value.id
     }
   }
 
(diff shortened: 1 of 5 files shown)


terraform/modules/paas/main.tf
```diff
@@ -76,18 +76,24 @@ resource cloudfoundry_app worker_app {
 }
 
 resource cloudfoundry_route web_app_cloudapps_digital_route {
+  count = var.disable_routes == null ? 1 : 0
```
saliceti, line 79: Do we need to move the cloudapps route?

RMcVelia: If they want to test a review app via the find route, then I'd think so. Also seems to be a risk to leave a route that would remain pointing to find.

terraform/variables.tf
```diff
@@ -34,6 +34,8 @@ variable key_vault_infra_secret_name {}
 
 variable azure_credentials { default = null }
 
+variable "disable_routes" { default = null }
```
saliceti, line 37: It would be more natural to make it a boolean. ie `true` instead of `"true"` and `false` instead of `null`

terraform/modules/paas/main.tf
saliceti: As discussed, this is required for review apps

terraform/variables.tf
RMcVelia: changed to a boolean

saliceti (approved):

PR #1699, opened 2023-01-25
950 update content

goodviber:
### Context
We need to update the content in Prod to match the prototype. We are changing the line of content that appears above the locations list in the 'School placement' section and the content that appears in the GIT 'Where you will train' box when the provider is a SCITT.

Because we are changing the content on Find, we'll also need to change it in the preview in Publish (see attached screenshot).

See attached card and let me know if you have any questions :)
Tech notes

    Make sure to have the counter work for new find in publish

### Checklist

- [x] Rebased `main`
- [x] Cleaned commit history


avinhurry:
[Change 'School placements' and 'Where you will train' content](https://trello.com/c/ivK8AKLO/950-change-school-placements-and-where-you-will-train-content)

avinhurry (commented):
Same for this one: https://github.com/DFE-Digital/publish-teacher-training/pull/3275#pullrequestreview-1269111303

avinhurry (approved):

spec/components/courses/about_schools_component_spec.rb
```diff
@@ -127,7 +127,7 @@
 
         result = render_inline(described_class.new(course))
 
-        expect(result.text).not_to include('You’ll be placed in different schools during your training. You can’t pick which schools you want to be in')
+        expect(result.text).not_to include("You'll be placed in different schools during your training. You can't pick which schools you want to be in")
```
defong, line 130:
```suggestion
        expect(result.text).not_to include("You’ll be placed in different schools during your training. You can’t pick which schools you want to be in")
```

https://smartquotesforsmartpeople.com/

